PR #177, opened 2015-05-14
Scroll viewpager fix

rfdy:
We added support for on scroll horizontal. This affects all view pagers to check if the tile view can scroll.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.1.0'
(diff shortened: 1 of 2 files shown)


moagrius:
nice, thanks

rfdy:
Thanks for the awesome library and the speedy merge. 

Would it be too much trouble to tag a new release and publish to maven?

moagrius:
It wouldn't be too much trouble, that said, I've never published to Maven - another contributor used to handle that...  I need to figure it out at some point, but that's not likely to be soon...  If you have the time, feel free to publish it if that's possible

PR #188, opened 2015-06-26
tile loading async

bnsantos:
I changed the way the tiles are loading, to user a thread pool. Also updated the disklrucache

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,14 +39,14 @@ dependencies {
 }
 
 task clearJar(type: Delete) {
-    delete 'build/outputs/tileviewlib-1.0.13.jar'
+    delete 'build/outputs/tileviewlib-1.0.14.jar'
 }
 
 task makeJar(type: Copy) {
     from('build/intermediates/bundles/release/')
     into('build/outputs/')
     include('classes.jar')
-    rename ('classes.jar', 'tileviewlib-1.0.13.jar')
+    rename ('classes.jar', 'tileviewlib-1.0.14.jar')
 }
 
 apply from: 'mvn-push.gradle'
--- a/build.gradle
+++ b/build.gradle
@@ -17,12 +17,12 @@ repositories {
 apply plugin: 'com.android.library'
 
 android {
-    compileSdkVersion 15
-    buildToolsVersion '19.1.0'
+    compileSdkVersion 22
+    buildToolsVersion '21.1.2'
 
     defaultConfig {
-        minSdkVersion 8
-        targetSdkVersion 15
+        minSdkVersion 15
+        targetSdkVersion 22
     }
 
     sourceSets {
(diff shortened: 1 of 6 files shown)


moagrius:
So after an very quick initial review, I like a lot of what you've done here; I'll have to dig into the thread stuff more to get a complete understanding, but one major issue is moving from min SDK of 8 to 15 - that's API 2.2 to API 4.0.3, which is a massively breaking change and there's no way I can merge that.  Can you achieve what you want to without changing the min SDK?  Also, try to keep with code style - I understand and appreciate the "m-" and "s-" prefixes, but since they're not used here, it wouldn't be appropriate - same with white space - you might have a completely justifiable reason for using the white space rules you do, but please conform to the existing set up.

bnsantos:
Totally agree with you, was in a rush. Going to check those items over the weekend!

moagrius:
Sounds good, looking forward to seeing the updates

bnsantos:
So I checked the things I'm using for async loading tiles (ThreadPoolExecutor, BlockingQueue, Runnable) all of them are available since API 1. Also formatted the code to match the current pattern!
Another stuff I updated is using appcompact-v7 but I don't think this is also going to be any problem supporting api beginning 8. To make sure I run a quick test on an emulator api 8 and everything worked fine.

moagrius:
Very cool - give me a few days to find some time to verify everything, but at first glance looks good

moagrius:
So I've reviewed (but haven't actually run it), and in general it looks good.  Can you explain the benefits of the changes you're proposing?  I see the value in the caching changes ("working" is always better than "not working"), but AsyncTask uses a ThreadPool behind the scenes anyway, so I'm not sold on why you moved to explicit ThreadPoolExecutor instance, but could be convinced...  Could you enumerate each change and it's material benefit?

Also, if we do merge, this'd require a version change (in both manifest and gradle) to 1.0.15

bnsantos:
Ok, I'll explain the issues I was facing using your lib before starting to work in a solution for that. Loading Tiles from s3 using BitmapDecoderHttp sometimes is too slow (especially on 3g here in Brazil). 

Why it is too slow the TileRenderTask (for line 36) was loading all Tiles that are visible in screen one at a time, so it would do a get request for one picture, download it, load it into to screen and the it will start the next Tile.

The solution I did was instead of using only one TileRenderTask to load ALL tiles I created one Runnable (TileDownloadRunner) to load one single Tile and using it with ThreadPoolExecutor android can handles multiples Runnables depending on the number of cores of the devices, also it could do some async stuff whenever I'm waiting for I/O operations it could load an already downloaded picture into screen which will looks more faster.

The main benefit was performance. Android uses ThreadPool behind the scenes but the current code was note creating multiple AsyncTasks to get the benefits of ThreadPool. I'm pretty sure I could dig into your code and find out a better solution using the same patterns you're using (AsyncTaks) to achieve the concurrent loading. But the solution I did was the fastest and simplest I could find right now.

moagrius:
I get it - sounds great.  I'll run it locally some time this weekend as a final sanity check then, assuming all's well, merge.  Thanks for the update and the walkthrough - I really appreciate it.

moagrius:
OK, I ran it, looks good - I've added you as a collaborator so you can merge yourself - a couple final requests for before/during merge:

1.  Could you update version to 1.0.15?
2.  Could you generate a JAR and add it to https://github.com/moagrius/TileView/releases?
3.  Could you push to maven and confirm availability?

LMK if you have any questions.

Thanks for contributing - very much appreciated.

bnsantos:
Ok,

I'll do 1 and 2, not sure how to do 3. What about creating a tag for 1.0.15???

bnsantos:
So the tag is created during the release process. It's just missing pushing jar to maven.

moagrius:
Hm, another collaborator used to do that - I thought it was part of the gradle process, but I'll double check later - anyway, thanks very much for the awesome contribution

Issue #205, opened 2015-10-19
Crashes with caching enabled


dmo60:
Hello @moagrius!

First of all, great library! It is exactly what I was looking for, so many features, thank you!

I want to use disk caching for my tiles. Your latest commits made it work quite beautifully, but I still have one issue with it. I am showing floor plans for university buildings and the user is able to switch the floor level. When he does so, I create a new TileView showing tiles for the new floor and destroy the old one (which was showing the previously selected floor). More or less, it looks like this:

```
        // new floor selected, destroy old tileview
        if (mTileView != null) {
            mTileView.setVisibility(View.GONE);
            mTileView.destroy();
            mTileViewContainer.removeView(mTileView);
        }

        // add new tileview
        mTileView = new TileView(getActivity());
        mTileView.setTileDecoder(new BitmapDecoderHttp());
        mTileView.setCacheEnabled(true);
        mTileView.setTransitionsEnabled(true);
        mTileView.addTileViewEventListener(this);

        mTileViewContainer.addView(mTileView);

        mTileView.setSize(...);
        // set detail levels, etc...
```

However, when switching floors, especially when doing so quite fast, it crashes regularly with the following exception:

```
java.lang.IllegalStateException
     at com.jakewharton.disklrucache.DiskLruCache.completeEdit(DiskLruCache.java:511)
     at com.jakewharton.disklrucache.DiskLruCache.access$2200(DiskLruCache.java:88)
     at com.jakewharton.disklrucache.DiskLruCache$Editor.abort(DiskLruCache.java:824)
     at com.qozix.tileview.tiles.TileCache.addBitmapToDiskCache(TileCache.java:158)
     at com.qozix.tileview.tiles.TileCache.addBitmap(TileCache.java:73)
     at com.qozix.tileview.tiles.Tile.decode(Tile.java:74)
     at com.qozix.tileview.tiles.TileManager.decodeIndividualTile(TileManager.java:302)
     at com.qozix.tileview.tiles.TileDownloadRunner.run(TileDownloadRunner.java:22)
     at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1113)
     at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:588)
     at java.lang.Thread.run(Thread.java:818)
```

I am using DiskLruCache v2.0.2.

[...]

PR #206, opened 2015-10-20
Catch all exceptions when caching fails

dmo60:
This fixed my issue as discussed here: #205 
Feel free to merge or reject :)


moagrius:
thank you!

PR #207, opened 2015-10-22
Add removeAllMarkers to MarkerManager

renemichel:
There was no option to remove all markers from the TileView's tree although there were
methods to remove individual markers, so references to the markers were stored thus making a
removeAllMarkers method not difficult and didn't need deep changes to the core code.


moagrius:
Looks good, thanks for the update.

FWIW, version 2 is coming out fairly soon, and is a major version bump so will immediately deprecate the current version (1).

PR #253, opened 2016-01-26
Remove AndroidManifest.xml defaults

ened:
These defaults might clash with a including project and cause lint errors unless the API client include `tools:replace="allowBackup,supportsRtl"` within their AndroidManifest.xml


moagrius:
Thank you!

Issue #256, opened 2016-02-03
Render during panning


peterLaurence:
Hey Mike,

This one is not an issue, it's just to ask you if you tried to make your TileView render tiles while panning/zooming. I was looking at your code today to start a custom implementation of parallel tile loading when i stumbled on how pan events are handled in TileView. I was surprised that TileView internally use RenderThrottleHandler, while i just expected render requests to be triggered by the end of a pan or a fling. I realized that the PanEnd event is fired at the end of a fling AND at the end of a pan.
I thought it would be simpler if the PanEnd is fired just once : at the end of the pan if there is no fling, or else at the end of the fling.

One modification leading to another.. i realized that we can render while paning/zooming without stutter.
You can have a look at my [repo](https://github.com/peterLaurence/TileView/commits/render-during-pan), branch "render-during-pan". But just to be clear : these are quick and dirty modifications. But it works, so i'd like you to know.

The behavior in the demo is amazing :) I don't even have anymore the issue #252.

moagrius:
interesting - i'll check it out - have you tried the version without debounce using tiles from a network source (e.g., http)?

peterLaurence:
No i haven't tried with network source, because i plan to implement asynchronous tile loading, as @bnsantos said. This implementation would be more suited for network source.

mkmckenna:
Hi Moagrius/Peter

I'm going to post this question here as its slightly related. Btw I have also done some customization of the panning and zooming logic so that it does this while doing these actions and it seems to work. Thats interesting though about it being called twice. I wonder if thats got something to do with some of the warnings I am seeing.

Anyway, I've recently moved to V2. I had a pretty big switch over rewrite to do for a number of reasons not really to do with TileView however.

One issue that I'm having though is I have a rendering task that is kicked off when TileView launches. This renderer basically builds loads of paths, draws them to a canvas and I make my own custom tiles out of it and overlay these over my base tile set and store them in cache to be used. Its all quite complicated but it works. 

What I'm having issues though with is when this task ends I try and call requestRender() but nothing seems to happen. My brain is slightly fried and its late so maybe im missing something here, but im sure in v1 this method worked correctly.

Btw V2 is a vast improvement. Brilliant work!

Cheers,
Martin

moagrius:
@peterLaurence 

> No i haven't tried with network source, because i plan to implement asynchronous tile loading, as @bnsantos said. This implementation would be more suited for network source.

[...]

PR #257, opened 2016-02-05
Add the ability to render while panning.

peterLaurence:
Like explained in #256.
I enabled this in the `LargeImageTileViewActivity` demo.

### DIFF
--- a/tileview/src/main/java/com/qozix/tileview/TileView.java
+++ b/tileview/src/main/java/com/qozix/tileview/TileView.java
@@ -82,6 +82,8 @@ public class TileView extends ZoomPanLayout implements
 
   private RenderThrottleHandler mRenderThrottleHandler;
 
+  private boolean mShouldRenderWhilePanning = false;
+
   /**
    * Constructor to use when creating a TileView from code.
    *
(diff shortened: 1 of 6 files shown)


tileview/src/main/java/com/qozix/tileview/TileView.java
```diff
@@ -704,6 +706,20 @@ public void resume() {
...
   }
 
+  /**
+   * Allows the TileView to render tiles while panning.
+   *
+   * @param shouldRender True if it should render while panning.
+   */
+  public void setShouldRenderWhilePanning( boolean shouldRender ) {
+    mShouldRenderWhilePanning = shouldRender;
+    if( shouldRender ) {
```
moagrius: when a condition just determines a value (rather than an "action" - think: method), i think ternary is appropriate:

```
int buffer = shouldRender ? TileCanvasViewGroup.FAST_RENDER_BUFFER : TileCanvasViewGroup.DEFAULT_RENDER_BUFFER;
mTileCanvasViewGroup.setRenderBuffer( buffer );
```

moagrius:
looks good!  +1 with one minor style change

thank you

moagrius:
also, can you update the readme (2 places) and the library's build.gradle to the next version (2.0.9)?

peterLaurence:
Ok done!

moagrius:
awesome, fast.  i'll merge as soon as i get a minute to generate the aars and upload to jcenter.  thanks again!

tileview/src/main/java/com/qozix/tileview/TileView.java
```diff
@@ -723,7 +736,11 @@ protected void updateViewport() {
   protected void onScrollChanged( int l, int t, int oldl, int oldt ) {
     super.onScrollChanged( l, t, oldl, oldt );
     updateViewport();
-    requestThrottledRender();
+    if( mShouldRenderWhilePanning ) {
+      requestRender();
```
moagrius: Just noticed - should this be `requestSafeRender`, in case it's flinging or programmatically panning (e.g., `slideTo`)?

peterLaurence: This is done in purpose, because if we call `requestSafeRender`, the fling will cause the call of `requestThrottledRender`, and we do not want this.
Even during a fling, there is no stutter while rendering.

moagrius: +1 (since it's optional anyway)

back to +1 on the whole PR, will merge when i get a minute

moagrius:
merged, and now available on jcenter :thumbsup:

peterLaurence:
nice :)

Issue #258, opened 2016-02-06
Optimize rendering task


peterLaurence:
I am creating the issue discussed earlier, just to keep track of it.

Rendering task is given the list of visible tiles, without taking into account the tiles already rendered. In fact, they are but each time the TileRenderTask adds a tile to the view. This is particularly visible when using tiles from a network : when the visible area overlaps an area already rendered, all the tiles already rendered are decoded/downloaded again.

@moagrius 
```
LinkedList<Tile> getRenderList() {
  LinkedList<Tile> renderList = (LinkedList<Tile>) mTilesScheduledToRender.clone();
  renderList.removeAll( mTilesAlreadyRendered );
  return renderList;
}
```
The snippet you proposed works perfectly. I am just checking if there is anything more to modify.

Issue #259, opened 2016-02-06
Unnecessary load of tiles from the highest scale `DetailLevel`


peterLaurence:
When double-tap is used at maximum scale to go back to the minimum scale, the tiles from the `DetailLevel` at maximum scale are rendered, instead of the ones from the current `DetailLevel` (after double-tap). 

To see that, just log as follows :
```
@Override
  public void onScaleChanged( float scale, float previous ) {
    super.onScaleChanged( scale, previous );
    mDetailLevelManager.setScale( scale );
    mHotSpotManager.setScale( scale );
    mTileCanvasViewGroup.setScale( scale );
    mScalingLayout.setScale( scale );
    mCompositePathView.setScale( scale );
    mMarkerLayout.setScale( scale );
    mCalloutLayout.setScale( scale );
    System.out.println( "scale : " + scale );
    System.out.println( "current scale : " + mDetailLevelManager.getCurrentDetailLevel().getScale() );
  }
```
Zoom in whatever demo to max scale an then double-tap. The tiles rendered afterwards have a "current scale" that is the same as before the double-tap. In other words, a lot of tiles may be requested to be rendered, especially on a map with many `DetailLevel`.

peterLaurence:
Removing the `DetailLevel` lock in `TileView` solves the issue. No jumpy behavior when zooming accross `DetailLevel` changes.

```
@Override
  public void onZoomBegin( float scale, Origination origin ) {
//    mDetailLevelManager.lockDetailLevel();
    mDetailLevelManager.setScale( scale );
  }
```
@moagrius 
But i guess you did that on purpose. Can you explain why you introduced that lock?

moagrius:
> But i guess you did that on purpose. Can you explain why you introduced that lock? Perhaps you were concerned that a DetailLevel change during a zoom (programmatic or not) would cause some issue?

Yes, definitely intentional.  Imagine someone was right at the edge of 2 detail levels (say .48 with a registered detail level of 0.25 and 0.5), and they started pinching, a little larger, a little smaller - without the lock, _all_ visible tiles are going to re-rendered _on top of_ the previous set, _each time there's a change_, so not only is that a ton of extra computration and a hit on performance, the likelihood of OOM is very high as well.

By locking the zoom level during a pinch, we ensure that the tiles that were visible when the pinch started scale, but are not swapped out _during the pinch_, although they may be replaced after the pinch is complete.

LMK if that doesn't make sense.

peterLaurence:
[...]

Issue #252, opened 2016-01-23
Sometimes, some tiles are not rendered


peterLaurence:
Hi,

Most of the time, tiles are correctly rendered. But sometimes, after an intense unzoom, some tiles are not rendered. At the beginning, i thought that my BitmapProvider was the culprit. Its implementation is very close to BitmapProviderAssets, because it looks for tiles located on the sd card. So i tried to log any Exception raised by BitmapFactory.decodeFile. But there were none.
At the moment, the only clue i have is that it happens just after a drop in memory usage of my app. I made a screen shot of that memory usage: 

![screen](https://cloud.githubusercontent.com/assets/15638794/12530271/c433df8c-c1d9-11e5-81ac-7df99e88002b.png)

As a side note, even if some tiles are not rendered, if i want them appear, i just have to scroll (or zoom / unzoom) a little to trigger tiles painting.

moagrius:
Does this ever happen in the demo to you, or just in your custom implementation?

You can try calling `invalidate` or `postInvalidate` on the `TileView`, or on the `TileCanvasViewGroup` returned by `TileVIew.getTileCanvasViewGroup`

peterLaurence:
Yes, it happens in the demo. Although it is not easy to trigger.
I chose the Large Image example (Joconde). What i do is :
• I zoom in the map to the maximum
• I unzoom out the map and i try to do it in one gesture (an "intense" unzoom). I don't make any gesture right after that.

After that, sometimes tiles are drawn correctly. But sometimes some of them are missing : 

![screenshot_20160124-113621](https://cloud.githubusercontent.com/assets/15638794/12535792/0f6579d8-c28f-11e5-89f5-ff856e3cc4a4.png)

To get back the missing tiles, we have to zoom in.

moagrius:
Hm, have you tried `invalidate/postInvalidate`, per https://github.com/moagrius/TileView/issues/252#issuecomment-174218710?

peterLaurence:
I don't understand why i should manually call invalidate/postInvalidate on the TileView object. It is not the upper layer's responsibility to tell the TileView to invalidate itself. It should do it automatically when it is necessary, which i think it does (i am not sure it is an invalidate issue).
I noticed that issue is most likely to happen with bigger maps than the ones in the demo. Right now, i am rendering maps with 6 DetailLevels, and i do severe tests with big zoom/unzoom, paning, ... 
[...]

PR #261, opened 2016-02-07
Fix issues #258, #252 and #259

peterLaurence:
Since the last PR i solved #252 and #259.

Solving #259 was not easy. While i was doing that i stumbled on the real culprit for #252 : in `TileCanvasViewGroup.beginRenderTask()`, after calling `mDetailLevelToRender.computeCurrentState()`, the method returned without starting a rendering task, because `computeCurrentState()` returned `false` in a situation where a render is really needed. E.g after a double tap from the maximum scale.

I tested all this in various situations, it works perfectly and it definitely improved the situation described in #259.

### DIFF
--- a/tileview/src/main/java/com/qozix/tileview/TileView.java
+++ b/tileview/src/main/java/com/qozix/tileview/TileView.java
@@ -772,7 +772,9 @@ public class TileView extends ZoomPanLayout implements
 
   @Override
   public void onZoomBegin( float scale, Origination origin ) {
-    mDetailLevelManager.lockDetailLevel();
+    if( !Origination.DOUBLE_TAP.equals( origin ) ) {
+      mDetailLevelManager.lockDetailLevel();
+    }
     mDetailLevelManager.setScale( scale );
   }
 
@@ -790,7 +792,6 @@ public class TileView extends ZoomPanLayout implements
 
   @Override
   public void onDetailLevelChanged( DetailLevel detailLevel ) {
-    requestRender();
     mTileCanvasViewGroup.updateTileSet( detailLevel );
   }
 
--- a/tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
+++ b/tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
@@ -31,6 +31,7 @@ public class TileCanvasViewGroup extends ScalingLayout implements TileCanvasView
   private TileRenderTask mLastRunTileRenderTask;
 
   private DetailLevel mDetailLevelToRender;
+  private DetailLevel mLastRequestedDetailLevel;
   private DetailLevel mLastRenderedDetailLevel;
   private TileCanvasView mCurrentTileCanvasView;
 
@@ -144,10 +145,10 @@ public class TileCanvasViewGroup extends ScalingLayout implements TileCanvasView
     if( mDetailLevelToRender == null ) {
       return;
     }
-    if( mDetailLevelToRender.equals( mLastRenderedDetailLevel ) ) {
+    if( mDetailLevelToRender.equals( mLastRequestedDetailLevel ) ) {
       return;
     }
-    mLastRenderedDetailLevel = mDetailLevelToRender;
+    mLastRequestedDetailLevel = mDetailLevelToRender;
     mCurrentTileCanvasView = getCurrentTileCanvasView();
     mCurrentTileCanvasView.bringToFront();
     cancelRender();
@@ -199,7 +200,7 @@ public class TileCanvasViewGroup extends ScalingLayout implements TileCanvasView
 
   private void beginRenderTask() {
     boolean changed = mDetailLevelToRender.computeCurrentState();
-    if( !changed ) {
+    if( !changed && mDetailLevelToRender.equals( mLastRenderedDetailLevel ) ) {
       return;
     }
     mTilesVisible = mDetailLevelToRender.getVisibleTilesFromLastViewportComputation();
@@ -248,8 +249,7 @@ public class TileCanvasViewGroup extends ScalingLayout implements TileCanvasView
     if( mTileRenderListener != null ) {
       mTileRenderListener.onRenderComplete();
     }
-    invalidate();
-    requestRender();
+    mLastRenderedDetailLevel = mDetailLevelToRender;
   }
 
   LinkedList<Tile> getRenderList() {
--- a/tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
+++ b/tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
@@ -22,7 +22,7 @@ public class TileCanvasViewGroup extends ScalingLayout implements TileCanvasView
   public static final int FAST_RENDER_BUFFER = 10;
   private static final int DEFAULT_TRANSITION_DURATION = 200;
 
-  private LinkedList<Tile> mTilesScheduledToRender = new LinkedList<Tile>();
+  private LinkedList<Tile> mTilesVisible = new LinkedList<Tile>();
   private LinkedList<Tile> mTilesAlreadyRendered = new LinkedList<Tile>();
 
   private BitmapProvider mBitmapProvider;
@@ -161,10 +161,10 @@ public class TileCanvasViewGroup extends ScalingLayout implements TileCanvasView
   public void clear() {
     suppressRender();
     cancelRender();
-    for( Tile tile : mTilesScheduledToRender ) {
+    for( Tile tile : mTilesVisible ) {
       tile.destroy( mShouldRecycleBitmaps );
     }
-    mTilesScheduledToRender.clear();
+    mTilesVisible.clear();
     for( Tile tile : mTilesAlreadyRendered ) {
       tile.destroy( mShouldRecycleBitmaps );
     }
@@ -202,7 +202,7 @@ public class TileCanvasViewGroup extends ScalingLayout implements TileCanvasView
     if( !changed ) {
       return;
     }
-    mTilesScheduledToRender = mDetailLevelToRender.getVisibleTilesFromLastViewportComputation();
+    mTilesVisible = mDetailLevelToRender.getVisibleTilesFromLastViewportComputation();
     if( mLastRunTileRenderTask != null && mLastRunTileRenderTask.getStatus() != AsyncTask.Status.FINISHED ) {
       mLastRunTileRenderTask.cancel( true );
     }
@@ -212,7 +212,7 @@ public class TileCanvasViewGroup extends ScalingLayout implements TileCanvasView
 
   private void cleanup() {
     LinkedList<Tile> condemned = new LinkedList<Tile>( mTilesAlreadyRendered );
-    condemned.removeAll( mTilesScheduledToRender );
+    condemned.removeAll( mTilesVisible );
     for( Tile tile : condemned ) {
       tile.destroy( mShouldRecycleBitmaps );
     }
@@ -253,7 +253,9 @@ public class TileCanvasViewGroup extends ScalingLayout implements TileCanvasView
   }
 
   LinkedList<Tile> getRenderList() {
-    return (LinkedList<Tile>) mTilesScheduledToRender.clone();
+    LinkedList<Tile> renderList = (LinkedList<Tile>) mTilesVisible.clone();
+    renderList.removeAll( mTilesAlreadyRendered );
+    return renderList;
   }
 
   void generateTileBitmap( Tile tile ) {
@@ -261,9 +263,6 @@ public class TileCanvasViewGroup extends ScalingLayout implements TileCanvasView
   }
 
   void addTileToCurrentTileCanvasView( Tile tile ) {
-    if( mTilesAlreadyRendered.contains( tile ) ) {
-      return;
-    }
     tile.setTransitionsEnabled( mTransitionsEnabled );
     tile.setTransitionDuration( mTransitionDuration );
     tile.stampTime();
--- a/tileview/src/main/java/com/qozix/tileview/widgets/ZoomPanLayout.java
+++ b/tileview/src/main/java/com/qozix/tileview/widgets/ZoomPanLayout.java
@@ -3,6 +3,7 @@ package com.qozix.tileview.widgets;
 import android.animation.Animator;
 import android.animation.ValueAnimator;
 import android.content.Context;
+import android.support.annotation.Nullable;
 import android.support.v4.view.ViewCompat;
 import android.util.AttributeSet;
 import android.view.GestureDetector;
@@ -379,15 +380,23 @@ public class ZoomPanLayout extends ViewGroup implements
    * @param focusX The horizontal focal point to maintain, relative to the screen (as supplied by MotionEvent.getX).
    * @param focusY The vertical focal point to maintain, relative to the screen (as supplied by MotionEvent.getY).
    * @param scale The final scale value the ZoomPanLayout should animate to.
+   * @param origination The optional origination of the scale change (can be null). Values from {@link com.qozix.tileview.widgets.ZoomPanLayout.ZoomPanListener.Origination}
    */
-  public void smoothScaleFromFocalPoint( int focusX, int focusY, float scale ) {
+  public void smoothScaleFromFocalPoint( int focusX, int focusY, float scale, @Nullable ZoomPanListener.Origination origination ) {
     scale = getConstrainedDestinationScale( scale );
     if( scale == mScale ) {
       return;
     }
     int x = getOffsetScrollXFromScale( focusX, scale, mScale );
     int y = getOffsetScrollYFromScale( focusY, scale, mScale );
-    getAnimator().animateZoomPan( x, y, scale );
+    getAnimator().animateZoomPan( x, y, scale, origination );
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public void smoothScaleFromFocalPoint( int focusX, int focusY, float scale) {
+    smoothScaleFromFocalPoint( focusX, focusY, scale, null );
   }
 
   /**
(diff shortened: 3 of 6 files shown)


tileview/src/main/java/com/qozix/tileview/TileView.java
```diff
@@ -790,7 +792,6 @@ public void onZoomEnd( float scale, Origination origin ) {
 
   @Override
   public void onDetailLevelChanged( DetailLevel detailLevel ) {
-    requestRender();
```
peterLaurence: Here, the requestRender is useless, since it is cancelled right away in `mTileCanvasViewGroup.updateTileSet( detailLevel );` and then called.

tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
```diff
@@ -22,7 +22,7 @@
   public static final int FAST_RENDER_BUFFER = 10;
   private static final int DEFAULT_TRANSITION_DURATION = 200;
 
-  private LinkedList<Tile> mTilesScheduledToRender = new LinkedList<Tile>();
+  private LinkedList<Tile> mTilesVisible = new LinkedList<Tile>();
```
peterLaurence: Rename

tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
```diff
@@ -248,12 +249,13 @@ void onRenderTaskPostExecute() {
     if( mTileRenderListener != null ) {
       mTileRenderListener.onRenderComplete();
     }
-    invalidate();
-    requestRender();
```
peterLaurence: That requestRender causes an infinite loop (`beginRender` being called, and it is definitely not necessary).

tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
```diff
@@ -261,9 +263,6 @@ void generateTileBitmap( Tile tile ) {
   }
 
   void addTileToCurrentTileCanvasView( Tile tile ) {
-    if( mTilesAlreadyRendered.contains( tile ) ) {
-      return;
-    }
```
peterLaurence: Not needed anymore

tileview/src/main/java/com/qozix/tileview/widgets/ZoomPanLayout.java
```diff
@@ -778,9 +794,10 @@ private boolean setupZoomAnimation( float scale ) {
       return false;
     }
 
-    public void animateZoomPan( int x, int y, float scale ) {
+    public void animateZoomPan( int x, int y, float scale, @Nullable ZoomPanListener.Origination origination ) {
```
peterLaurence: We can now produce various behaviors depending on the origination of the zoom.

tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
moagrius: This is in case there was any changes to the tiles to be displayed since the render process started.    If there is no change, it'll fast-fail out anyway.

tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
moagrius: Assuming you're saying it's not needed anymore because we remove them all later, I think this is still a good idea to keep, so there's less traversal when removal happens (and, semantically, the tile shouldn't be added if it's already rendered).  Finally, the more atomic we can keep each PR the better - if we want to optimize later, no problem, but let's make sure we're only touching moving parts we're 100% sure about.

tileview/src/main/java/com/qozix/tileview/widgets/ZoomPanLayout.java
moagrius: this is a breaking change.  that's a public method.  you can add a second signature like you've described, but we can't have any breaking changes.

tileview/src/main/java/com/qozix/tileview/TileView.java
```diff
@@ -772,7 +772,9 @@ public void onPanEnd( int x, int y, Origination origin ) {
 
   @Override
   public void onZoomBegin( float scale, Origination origin ) {
-    mDetailLevelManager.lockDetailLevel();
+    if( !Origination.DOUBLE_TAP.equals( origin ) ) {
```
moagrius: any zoom should not change detail levels while zooming.  it can change detail levels after, but not during - thus the lock here.  what's the rationale in doing this?

tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
```diff
@@ -31,6 +31,7 @@
   private TileRenderTask mLastRunTileRenderTask;
 
   private DetailLevel mDetailLevelToRender;
+  private DetailLevel mLastRequestedDetailLevel;
```
moagrius: what's the rationale behind separating `mLastRequestedDetailLevel ` and `mLastRenderedDetailLevel`?

tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
moagrius: hm after looking through more of the changes, i think my comment above https://github.com/moagrius/TileView/pull/261/files#r52124981 is probably not relevant.  what's the rationale here?

tileview/src/main/java/com/qozix/tileview/widgets/ZoomPanLayout.java
```diff
@@ -379,15 +380,23 @@ public void smoothScaleTo( float destination ) {
    * @param focusX The horizontal focal point to maintain, relative to the screen (as supplied by MotionEvent.getX).
    * @param focusY The vertical focal point to maintain, relative to the screen (as supplied by MotionEvent.getY).
    * @param scale The final scale value the ZoomPanLayout should animate to.
+   * @param origination The optional origination of the scale change (can be null). Values from {@link com.qozix.tileview.widgets.ZoomPanLayout.ZoomPanListener.Origination}
    */
-  public void smoothScaleFromFocalPoint( int focusX, int focusY, float scale ) {
+  public void smoothScaleFromFocalPoint( int focusX, int focusY, float scale, @Nullable ZoomPanListener.Origination origination ) {
```
moagrius: breaking change on a public method.  also, what's the rationale?

moagrius:
first, big thanks for digging in and working up fixes for the issues you've described.  that said, it feels like we're touching a lot of things in this PR, several of which aren't directly related to the issues described.  i'm not a big fan of bolting-on some functionality to some simple paradigms, e.g., `broadcastDoubleTapZoomBegin`.  i'm less concerned about adding `DOUBLE_TAP` to the `Origination` enum (since it already has gestures listed, this feels fairly reasonable), but i'm not sure why we're singling double-tap out anyway - i can't think of any reason a double-tap scale action would be treated any differently than any programmatic zoom action.  has this whole bit been carefully considered?

if a public method signature changes, it's considered a breaking change - imagine there's some build out there on 2.0.4 that has this line somewhere `mTileView.animateZoomPan( 0, 0, 0.5 );`.  if they went to upgrade to 2.0.10, their app would fail to compile, since there's now no signature matching what they've got implemented.  you can always _overload_ a signature, or add a new one, but you can't change any public api.  LMK if that doesn't make sense.

i'm not totally unwilling to consider this PR (minus the breaking changes), if you're convinced that all the changes are entirely justified, and want to explain, but it might be easier to back out some of the non-critical stuff and touch as little as possible.  i'm happy to help in any way i can, or collaborate, although i'm totally confident you can handle it on your own if you want to.  how to move forward at this point is totally up to you.

tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
peterLaurence: No we do not remove them later, at this point we already removed them so `mTilesAlreadyRendered` can't contain any tiles passed by `addTileToCurrentTileCanvasView`.

That was part of the previous PR.

tileview/src/main/java/com/qozix/tileview/widgets/ZoomPanLayout.java
peterLaurence: That's why i added :
```
/**
   * {@inheritDoc}
   */
  public void smoothScaleFromFocalPoint( int focusX, int focusY, float scale) {
    smoothScaleFromFocalPoint( focusX, focusY, scale, null );
  }
```
Not sure though if my @inheritDoc is at a good place.

tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
peterLaurence: While it is indeed fast-fail out, I believe that if a new render is need since a change to the tiles to be displayed, this is the responsibility of `TileView`. For example, we have a `requestRender` at the end of a pan, at a `DetailLevel`change, etc.
Letting a `requestRender` here just causes permanent extra computation with ` DetailLevel.computeCurrentState()`. I don't know the impact on cpu usage (and thus on battery life), but if that can be avoided, why not do it?

tileview/src/main/java/com/qozix/tileview/widgets/ZoomPanLayout.java
peterLaurence: I also added :
```
public void animateZoomPan( int x, int y, float scale ) {
  animateZoomPan( x, y, scale, null );
}
```

peterLaurence:
>  i'm less concerned about adding DOUBLE_TAP to the Origination enum (since it already has gestures listed, this feels fairly reasonable), but i'm not sure why we're singling double-tap out anyway - i can't think of any reason a double-tap scale action would be treated any differently than any programmatic zoom action. has this whole bit been carefully considered?

In fact, i don't like either the idea of singling double-tap out. I did that because i wanted to keep the lock at a zoom start, expect for a double-tap. But indeed, nothing prevents the misuse of `TileView` and do a programmatic zoom from max scale to min scale. You talked the other day of the likelyhood of OOM. This is a real one (i am working with much bigger maps than in the demo, and #259 is critical). Imagine you zoom out from scale 2^8 (8 `DetailLevels` to scale 0.5), with `TileView.setShouldScaleToFit( false )`. You will try to render a LOT of tiles.
So how do we handle that? That's not easy. I found that removing the lock was a lazy way while NOT having the risk of OOM, because even if a zoom causes very quick `DetailLevel` changes, it calls `requestRender`, which by nature limits the number of rendering operations, thanks to the `TileCanvasViewGroup.TileRenderHandler` (remember `DEFAULT_RENDER_BUFFER = 250`).
And even with `FAST_RENDER_BUFFER`, there is no issue.

tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
peterLaurence: While i allowed a zoom to trigger `DetailLevel` changes (just in case of a double-tap), i finally understood what caused #252. I warn you : this is complicated.

So you need a zoom that will cause a render of a lot of tiles. Double-tap from max scale is perfect. You need a map with enough `DetailLevel`. The demo may be enough. To emphasize this, set `TileView.setShouldScaleToFit( false )` and have a minimum scale equal or below the minimum `DetailLevel`'s scale to be be able to see every levels (and trigger the issue). So from max scale you double-tap. What happens next? Because of the lock, the `DetailLevel` don't change until the zoom ends. But before it ends, numerous tiles from that level will be rendered and scaled because during the zoom, `TileView.onScrollChanged` is being called (which indirectly calls `requestRender`). That's were we have a likelyhood of OOM. When the zoom ends, it removes the lock, updates the scale of the `DetailLevelManager`, which in turn calls `DetailLevelManager.update()` and then triggers a `DetailLevel` change. Given the impelmentation of a `DetailLevel` change in `TileView` :
```
@Override
public void onDetailLevelChanged( DetailLevel detailLevel ) {
   requestRender();
   mTileCanvasViewGroup.updateTileSet( detailLevel );
  }
```
the updateTileSet calls a `cancelRender`. So it attempts to stop the ongoing task mentioned before (which takes some time because of the number of the tiles to render). At this point we may say : "that's ok, after that a `requestRender` is called and everything will be fine". BUT, it turn out, that request may just have a `DetailLevel.computeCurrentState()` returning false, so the `requestRender` does nothing. So why `computeCurrentState` may return false? While zooming, and ongoing `requestRender` may call `computeCurrentState` at a point were the current `DetailLevel` is equal to the "destination" one (idk how to say this :). So, in `DetailLevel`, the attribute `mLastStateSnapshot` already has its end value, and so as the other parameters `rowStart`, `columnStart`, ...
And there we have it : the requestRender that does nothing! :)

tileview/src/main/java/com/qozix/tileview/TileView.java
peterLaurence: have you read my comment in #259, saying that changing detail level while zooming do not cause OOM?
If you think about it, if we allow rendering while panning, why should we forbid rendering while zooming?

moagrius: > if we allow rendering while panning, why should we forbid rendering while zooming?

a couple important points: 1, we allow render _optionally_, and it is not default behavior, the user must opt-in to render while panning, while this change affects all users and all situations.  2., `lockDetailLevel` does not prevent rendering, it prevents the detail level from changing.

tileview/src/main/java/com/qozix/tileview/TileView.java
moagrius: so it's technically not useless - it's cancelled (and re-requested) in `updateTileSet` _only_ if the `DetailLevel` changes.  i'm a little hesitant to remove this - seems a little out of scope for the PR.

tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
moagrius: great explanation, and good point about the possibility of having to render a ton of tiles if scaled down to a point where all tiles are visible.  it'd be even worse if the user had only a single, very large detail level, and forced all those tiles to be rendered at once - it'd definitely OOM (that'd fail, silently, since decode OOM are caught, but still - there'd be missing pieces).  that said, i think that it's the user's responsibility if he allows that to happen - we can make reasonable attempts to be as responsible as possible, but we can't avoid irresponsible use-cases.

anyway, i'm fine with the change here (adding `mLastRequestedDetailLevel`) - it sounds like a reasonable way to handle the edge case you're describing.

tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
moagrius: i'd push back here - mostly because i can imagine a scenario where a request render happens, then state is changed in a subtle way that does not start a new render request, but also because it seems a bit out of scope here (more on that later).

tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
moagrius: hm i don't follow exactly but i believe you.  if you're 100% confident there's no chance of that test ever being true, +1 on this line

tileview/src/main/java/com/qozix/tileview/widgets/ZoomPanLayout.java
moagrius: you're totally correct, my mistake, i just saw in the diff that the original signature had a - next to it, and was replaced by your new signature, but when i look at the complete file i do see that they are new signatures, and the old signature pipes through, so it is _not_ a breaking change.  that said, not sure i love passing origination around, especially if it's null, seems like it might be a little misleading - but, i don't hate it either - more on overall approach later.

tileview/src/main/java/com/qozix/tileview/widgets/ZoomPanLayout.java
moagrius: you're correct, https://github.com/moagrius/TileView/pull/261/files#r52179298 applies here as well

moagrius:
i've commented the line notes.  generally speaking, i think the most important thing is: you identified a problem where the wrong detail level was being rendered, and you have a fix for that. 

i think anything other than what's needed to accomplish that should **not** be included in this PR.  i'd suggest what we respect the scope of that (or any scope directly related to issues being addressed by the PR), and not touch anything outside of it  - we can definitely optimize other things in other, later PRs, but let's keep this PR as atomic as possible.

i'll preface this next with: OOM is bad.  we can all agree on that.  however, if the widget user (the programmer, not the end-user) abuses the library, or makes bad engineering decisions, he should be allowed to fail.  i'm much more concerned about _incorrect_ behavior: you described that a specific case where a change in zoom rendered the wrong detail level.  that absolutely is a bug and absolutely needs corrected.  i believe this specific issue should be considered very separately from edge-cases that might result in OOM.

> I am feeling that handling that in the right way (by still permitting rendering while zooming) will break some mechanism in place. You don't like that, do you :)

remember that `lockDetailLevel` does not prevent render while zooming, it prevents detail level change.  i don't mind if it renders while zooming, if the user has opted in (`setShouldRenderWhilePanning`).  i believe that all zoom operations should prevent detail level from changing until the zoom is complete.  i'm not dogmatic about that, and could be convinced otherwise, but i think that the possibility of a new detail level rendering on top of an old detail level, during a zoom operation, is badly engineered - even if memory were no concern, just the UX of the tile set swapping depths, perhaps incompletely rendered, back and forth, strikes me as terrible.

thanks for all your work so far!

tileview/src/main/java/com/qozix/tileview/TileView.java
peterLaurence: My bad i wasn't clear, i wanted to say "why should we forbid detail level change".

peterLaurence:
Ok so i get it. More atomic PR on the way. I am just concerned that our conversation gets lost if i close this PR.

moagrius:
> what about preventing rendering operation while zooming out?

sounds reasonable :thumbsup:

re: everything else: cool, thank you sir!

peterLaurence:
Glad we found a settle :+1:

Issue #250, opened 2016-01-20
Loading more than one tile async


bnsantos:
Hey,
I did a PR while ago when the lib was still in version 1.something to fetch more than one tile at the same time. And it seems that this code was not introduced in version 2.0, all tiles are loaded in sequence (takes too long to load all tiles if they're not as a resource inside the app). 

Do you have interest in updating your lib to load more than one tile at the same time? I could possibly work on that and create another PR but before I would like to know why do you remove the code I did for this (was there a bug) not in the scope of the lib?

Thanks,

moagrius:
Hi - I remember the update.  I don't remember specifically how the final threading model was decided on, but virtually _everything_ in version 2 was reviewed and rewritten.  I would guess it's that the current iteration uses a `ThreadPoolExecutor`, so it's not really one at a time, it's just ordered - I believe that it'll start as many operations as there are cores available, but yes they'd be executed in order (and might potentially block subsequent ops; I'd have to look it up and am only half-in right now (still at work)).

Do you believe your approach would demonstrat better actual (versus perceptual) performance?  Can you briefly describe (or remind me) what your alternate approach was?  Unfortunately I squashed and lost history with v2.

bnsantos:
Hey,

I'm just getting familiar with your whole new code. But it seems that is exactly the same point as the previous versoin. You have a ThreadPoolExecutor that usually only have one async task which is responsible for loading all tiles sequentially (TileRenderTask in TileCanvasViewGroup.java). 

A faster approach would be having each tile render in a separate Thread this way the ThreadPoolExecutor could run more than one Thread at the same time. But in a approach like that we need to have a better way to decide if the task should run or not, because every time you scroll or do something else you should cancel all previous queued tasks and queue new tasks for each tile (you just need to cancel one task in your code right now and create a new instance of TileRenderTask).

I worked in this multi thread approach because I'm using the TileView with assets on the Internet, being able to download more than one at the same time is a must. I could possibly work in a way to make Tile loading multi threaded and do a PR for you to review.

It seems that my fork is still in the same state as the PR I did, if you want to take a look
https://github.com/bnsantos/TileView

moagrius:
Thanks for the link - that does look familiar.

The `AsyncTask` for fetching tile bitmaps in v1 was _not_ using `executeOnExecutor(THREAD_POOL_EXECUTOR)`, but v2 does.  

[...]

PR #263, opened 2016-02-16
Multithread tile rendering

bnsantos:
Improvements of this PR were discussed on 
https://github.com/moagrius/TileView/issues/250


tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
```diff
@@ -262,12 +268,18 @@ void generateTileBitmap( Tile tile ) {
...
+    mTileRenderHandler.post( new Runnable() {
+      @Override
+      public void run() {
+        tile.setTransitionsEnabled( mTransitionsEnabled );
+        tile.setTransitionDuration( mTransitionDuration );
+        tile.stampTime();
+        mTilesAlreadyRendered.add( tile );
+        mCurrentTileCanvasView.addTile( tile );
+      }
+    });
```
peterLaurence: little space missing here ;)

moagrius:
So I've had a quick look and I'd say this looks _really_ good.  There are a couple questions and fine points, but I'm excited to get this included.  Will start line comments now.

moagrius:
ah i just saw @peterLaurence has started - @peterLaurence let me know when you're done and I'll start, so we don't bump into each other.

thanks both

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.LinkedList;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.Future;
+import java.util.concurrent.LinkedBlockingDeque;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+public class TileRenderPoolExecutor {
+  // Sets the amount of time an idle thread will wait for a task before terminating
+  private static final int KEEP_ALIVE_TIME = 5;
```
peterLaurence: So we have a KKEP_ALIVE_TIME of 5s.
Have you found that this is better than the former 1s?

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
```diff
@@ -0,0 +1,115 @@
...
+  private class CustomThreadPoolExecutor extends ThreadPoolExecutor{
+    public CustomThreadPoolExecutor( int corePoolSize, int maximumPoolSize, long keepAliveTime, TimeUnit unit, BlockingQueue<Runnable> workQueue ) {
+      super( corePoolSize, maximumPoolSize, keepAliveTime, unit, workQueue );
+    }
+
+    @Override
+    protected void afterExecute( Runnable r, Throwable t ) {
+      super.afterExecute( r, t );
+//      getActiveCount() == 1 to make sure it is going to execute only for the last thread to run
+      if( mQueue != null && mQueue.size() == 0 && mViewGroup != null && mViewGroup.get() != null && getActiveCount() == 1) {
```
peterLaurence: Ok. I guess this called for each `TileRenderRunnable` created?

peterLaurence:
@moagrius 

Oups sorry. I am done :)

tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
```diff
@@ -242,14 +243,19 @@ void onRenderTaskCancelled() {
...
     mIsRendering = false;
-    if( !mTransitionsEnabled ) {
-      cleanup();
-    }
-    if( mTileRenderListener != null ) {
-      mTileRenderListener.onRenderComplete();
-    }
-    invalidate();
-    requestRender();
+    mTileRenderHandler.post( new Runnable() {
```
moagrius: question: have you found a benefit to using throw-away anonymous instances of things like `Runnable` preferable to either A) creating a single instance as a member variable and passing that instance to `.post`, or B) implementing `Runnable` and passing bare instances e.g., `.post( new RenderTaskPostExecuteRunnable() );`

i'm not saying it has to be one way or another, or even different than what you've done - just asking your opinion.  i personally believe it to be most efficient (in this case) to use alternative A - define the `Runnable` once as a member variable of `TileCanvasViewGroup` and pass that to `Handler.post` each time, but am definitely open to other ways if there's a reason to.

@peterLaurence feel free to weigh in here too

tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
```diff
@@ -262,12 +268,18 @@ void generateTileBitmap( Tile tile ) {
...
   }
 
-  void addTileToCurrentTileCanvasView( Tile tile ) {
-    tile.setTransitionsEnabled( mTransitionsEnabled );
-    tile.setTransitionDuration( mTransitionDuration );
-    tile.stampTime();
-    mTilesAlreadyRendered.add( tile );
-    mCurrentTileCanvasView.addTile( tile );
+  void addTileToCurrentTileCanvasView( final Tile tile ) {
+    mTileRenderHandler.post( new Runnable() {
```
moagrius: same basic question as https://github.com/moagrius/TileView/pull/263/files#r53020670, but in this case it'd have to be alternative B, since we'd need to attach the `Tile` reference.

again, not insisting on a change, just want to discuss and see what the best way is.

tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
```diff
@@ -262,12 +268,18 @@ void generateTileBitmap( Tile tile ) {
...
+      @Override
+      public void run() {
+        tile.setTransitionsEnabled( mTransitionsEnabled );
+        tile.setTransitionDuration( mTransitionDuration );
+        tile.stampTime();
+        mTilesAlreadyRendered.add( tile );
+        mCurrentTileCanvasView.addTile( tile );
```
moagrius: would lines 275 through 279 make sense as an instance method of `TileCanvasViewGroup`?  e.g., `prepateTileForRender` or something?

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
```diff
@@ -0,0 +1,115 @@
...
+  private WeakReference<TileCanvasViewGroup> mViewGroup;
+
+  private static TileRenderPoolExecutor sInstance = null;
+
+  static {
+    KEEP_ALIVE_TIME_UNIT = TimeUnit.SECONDS;
+    sInstance = new TileRenderPoolExecutor();
+  }
+
+  public TileRenderPoolExecutor() {
```
moagrius: shouldn't this be `private`?

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
```diff
@@ -0,0 +1,115 @@
...
+  // Sets the amount of time an idle thread will wait for a task before terminating
+  private static final int KEEP_ALIVE_TIME = 5;
+  private static final TimeUnit KEEP_ALIVE_TIME_UNIT;
+
+  private static final int CORE_POOL_SIZE = Runtime.getRuntime().availableProcessors();
+  private static final int MAXIMUM_POOL_SIZE = Runtime.getRuntime().availableProcessors();
+
+  private final CustomThreadPoolExecutor mExecutor;
+  private final BlockingQueue<Runnable> mQueue;
+  private final BlockingQueue<Future> mFutureList;
```
moagrius: we're saving them in the future list because we need to be able to cancel them later - is that correct?

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
```diff
@@ -0,0 +1,115 @@
...
+  private WeakReference<TileCanvasViewGroup> mViewGroup;
+
+  private static TileRenderPoolExecutor sInstance = null;
+
+  static {
+    KEEP_ALIVE_TIME_UNIT = TimeUnit.SECONDS;
```
moagrius: this doesn't need to be in a static block, you could just say back on line 14 `private static final TimeUnit KEEP_ALIVE_TIME_UNIT = TimeUnit.SECONDS;`

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
```diff
@@ -0,0 +1,115 @@
...
+    sInstance = new TileRenderPoolExecutor();
```
moagrius: I personally prefer the "on demand" singleton: https://en.wikipedia.org/wiki/Singleton_pattern#Initialization-on-demand_holder_idiom, but again am not dogmatic about it.  would you say the idiom you're using is more suitable for certain situations?

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
```diff
@@ -0,0 +1,115 @@
...
+      mFutureList.clear();
+    }
+  }
+
+  public void queue( TileCanvasViewGroup viewGroup, LinkedList<Tile> tiles ) {
+    if( tiles != null && tiles.size() > 0) {
+      mViewGroup = new WeakReference<>( viewGroup );
+      viewGroup.onRenderTaskPreExecute();
+      int size = tiles.size();
+      for( int i = 0 ; i < size ; i++ ) {
```
moagrius: any reason you're not using `for( Title tile : tiles ) {` here?  I believe with a `LinkedList` it's a more efficient iteration (since `LinkedList.get(index)` is O(n)).

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
```diff
@@ -0,0 +1,115 @@
...
+      }
+    }
+  }
+
+  static class TileRenderRunnable implements Runnable {
+    private final WeakReference<TileCanvasViewGroup> mTileCanvasViewGroup;
+    private final WeakReference<Tile> mTile;
+
+    public TileRenderRunnable( TileCanvasViewGroup viewGroup, Tile tile) {
+      this.mTileCanvasViewGroup = new WeakReference<>( viewGroup );
```
moagrius: please remove `this.` on this line and the next

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
```diff
@@ -0,0 +1,115 @@
...
+      this.mTile = new WeakReference<>( tile );
+    }
+
+    @Override
+    public void run() {
+      android.os.Process.setThreadPriority( android.os.Process.THREAD_PRIORITY_BACKGROUND );
```
moagrius: is there a reason for the FQNS here, rather than `import android.os.Process; ... Process.setThreadPriority(...`?

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
```diff
@@ -0,0 +1,115 @@
...
+    public CustomThreadPoolExecutor( int corePoolSize, int maximumPoolSize, long keepAliveTime, TimeUnit unit, BlockingQueue<Runnable> workQueue ) {
+      super( corePoolSize, maximumPoolSize, keepAliveTime, unit, workQueue );
+    }
+
+    @Override
+    protected void afterExecute( Runnable r, Throwable t ) {
+      super.afterExecute( r, t );
+//      getActiveCount() == 1 to make sure it is going to execute only for the last thread to run
+      if( mQueue != null && mQueue.size() == 0 && mViewGroup != null && mViewGroup.get() != null && getActiveCount() == 1) {
+        mViewGroup.get().onRenderTaskPostExecute();
```
moagrius: don't we need a null test here, in case the `WeakReference` was released?

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
```diff
@@ -0,0 +1,115 @@
...
+      this.mTile = new WeakReference<>( tile );
+    }
+
+    @Override
+    public void run() {
+      android.os.Process.setThreadPriority( android.os.Process.THREAD_PRIORITY_BACKGROUND );
+      final Thread thread = Thread.currentThread();
+
+      TileCanvasViewGroup tileCanvasViewGroup = mTileCanvasViewGroup.get();
+      Tile tile = mTile.get();
```
moagrius: nit: this should probably be moved to line 86, since if we fast-fail out on line 84, there's no reason for the fetch.  total nitpick, feel free to ignore me :)

moagrius:
@bnsantos i think that's it from me.  despite the number of comments, i really like this update.  i'm not as familiar with the multithreading as i should be so i'm glad someone who is was able to write this.

@peterLaurence feel free to weigh in on any of my comments, and thanks for helping

tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
peterLaurence: We are reaching this portion of code from multiple threads (not from ui thread).
From doc of the post method :

> Causes the Runnable r to be added to the message queue. The runnable will be run on the thread to which this handler is attached.

In this case, the handler is attached to the ui thread. So i would say that it's ok to post the same instance of `Runnable`, because then no mutual exclusion is need to protect us from accessing ui object from multiple threads.

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
bnsantos: Yeah the future list is just to be able to cancel kind of during the Runnable is being executed.

moagrius: awesome, thanks for clarifying

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
bnsantos: I changed for testing, forgot it

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
bnsantos: yes, this afterExecute is a hook that is called every time a Runnable is done (but its state is still running) and about to finish

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
peterLaurence: Yes, that seemed a bit long :)

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
bnsantos: sure

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
peterLaurence: ty for clarification

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
bnsantos: I can change it, I just copied that code from somewhere else.

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
bnsantos: Since I watched [this](https://www.youtube.com/watch?v=b6zKBZcg5fk) I stopped using the for like `for( Title tile : tiles )`. But if you prefer to keep a pattern in your code base I can change it.

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
bnsantos: there is already a test for `mViewGroup != null && mViewGroup.get() != null`. Or is it about anything else?

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
bnsantos: copy and paste. No reason at all

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
moagrius: if you're using an array-backed list, then you're correct, the pattern you're showing is more efficient.  however, a LinkedList has random access at O(n), because it does not have an array behind it.  I just reviewed the video you linked, and he even states very emphatically that you should *not* do this on a LinkedList (near the very end of the video).

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
moagrius: ah you're correct, my mistake, please disregard

tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
bnsantos: Not sure, they're not used in any other part of the code.

moagrius: not a big deal.  semantically it makes sense, but i can always tweak that kind of thing later.  either way - up to you

tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
bnsantos: One of the benefits of using B is that we can cancel it during execution (if required, but I can't see any case that this could be used). By performance I don't know which one is more efficient. 

Now considering the `addTileToCurrentTileCanvasView` case. I believe that we need to create new instances of runnable exactly because of the Tile passed as parameter (this method is called by every tile not only once per rendering like the `onRenderTaskPostExecute`

moagrius: IIUC, sounds like we're all on the same page.  So: A in the first example, and B in the second example?

tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
bnsantos: Created a `PrepareTileForRenderRunnable` class and a method `prepareTileForRender` on `TileViewGroup`for that. I believe solves both "issues" regarding the same peace of code

bnsantos:
@moagrius and @peterLaurence 
I believe I did all the changes required. If I skipped one please let me know

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
```diff
@@ -0,0 +1,113 @@
...
+
+  private static TileRenderPoolExecutor sInstance = null;
+
+  private TileRenderPoolExecutor() {
+    mQueue = new LinkedBlockingDeque<>();
+    mExecutor = new CustomThreadPoolExecutor( CORE_POOL_SIZE, MAXIMUM_POOL_SIZE, KEEP_ALIVE_TIME, KEEP_ALIVE_TIME_UNIT, mQueue );
+    mFutureList = new LinkedBlockingDeque<>();
+  }
+
+  public static TileRenderPoolExecutor getsInstance() {
```
moagrius: So can we use the on-demand idiom i linked to previous: https://en.wikipedia.org/wiki/Singleton_pattern#Initialization-on-demand_holder_idiom - something like:

```
private static class SingletonHolder {
  private static final TileRenderPoolExecutor INSTANCE = new TileRenderPoolExecutor();
}
public static TileRenderPoolExecutor getInstance() {
  return SingletonHolder.INSTANCE;
}
```

That said, I don't see any reason why this needs to be a singleton at all - if you want to just instantiate an instance in `TileCanvasViewGroup` constructor, that'd be fine as well.

moagrius:
@bnsantos just one last thing (the singleton idiom) and +1 from me

tileview/src/main/java/com/qozix/tileview/tiles/TileRenderPoolExecutor.java
bnsantos: ok, not going to use singleton anymore. Instantiated on `TileCanvasViewGroup` constructor

moagrius:
looks great, thank you.  +1 from me.  @peterLaurence are you +1?  if so I'll merge and generate artifacts for jcenter in about 2 hours (after lunch and scrum).

thanks again @bnsantos

peterLaurence:
+1 too

moagrius:
awesome, merging

bnsantos:
thanks guys!

moagrius:
thank _you_!  working on getting artifacts up to bintray and bumping version, will do as soon as i get a second to breathe

moagrius:
@bnsantos i merged and pulled down master to test and generate artifacts.  a couple things, most important: i'm getting `java.io.InterruptedIOException` pretty regularly (and improperly drawn bitmaps) on the new demo.  are you seeing that?  also, it appears to run from top left to bottom right, in order, which implies to me that the blocking queue is blocking until the previous task is done...  i remember in the previous version they appeared kind of randomly, which meant to me that they appeared as soon as they were done downloading, so truly parallel...  am i misunderstanding anything, or did we lose (effective) parallelism with the blocking queue?

bnsantos:
I'm seeing a lot of `java.io.InterruptedIOException` but this is because of a shitty implementation of `BitmapHttpProvider`. That happens whenever you pan waits is load a little bit and pan again then it will interrupt the executing Thread and then it throws `java.io.InterruptedIOException`.

The thread can be executed in any order (they're queued from 0_0 to row_col), I would say it's just coincidence that the tiles loaded pretty much like it was sequentially (top left to bottom right). 

The `ThreadPoolExecutor` can render more than one tile at the same time if so you could see a pattern of two tiles being loaded pretty much at the same time. Or sometimes one tile still loading between them. Or even some weird scenario. 

I'll take a look into the wrong bitmap loading.

moagrius:
> I'm seeing a lot of java.io.InterruptedIOException but this is because of a shitty implementation of BitmapHttpProvider. That happens whenever you pan waits is load a little bit and pan again then it will interrupt the executing Thread and then it throws java.io.InterruptedIOException.

hopefully we can get that fixed before publishing

> The thread can be executed in any order (they're queued from 0_0 to row_col), I would say it's just coincidence that the tiles loaded pretty much like it was sequentially (top left to bottom right).

> The ThreadPoolExecutor can render more than one tile at the same time if so you could see a pattern of two tiles being loaded pretty much at the same time. Or sometimes one tile still loading between them. Or even some weird scenario.

GTK.  i remember you'd done a video at one point that showed the difference in render times - do you feel like we're still getting all the benefits of that version?

> I'll take a look into the wrong bitmap loading.

cool, thanks.  fwiw, it's not the wrong bitmap, it's just a "blank" (it has pixels, but generally all white) - maybe the connection is dropped before it's fully decoded, but it's still reporting as successful?

here's some stacktrace that might be helpful:

```
02-16 15:09:16.474 19490-19600/tileview.demo D/skia: ---- read threw an exception
02-16 15:09:16.474 19490-19597/tileview.demo W/System.err:     at com.android.okio.RealBufferedSource.read(RealBufferedSource.java:48)
02-16 15:09:16.474 19490-19597/tileview.demo W/System.err:     at com.android.okhttp.internal.http.HttpConnection$FixedLengthSource.read(HttpConnection.java:442)
02-16 15:09:16.474 19490-19597/tileview.demo W/System.err:     at com.android.okio.RealBufferedSource$1.read(RealBufferedSource.java:168)
02-16 15:09:16.474 19490-19597/tileview.demo W/System.err:     at android.graphics.BitmapFactory.nativeDecodeStream(Native Method)
02-16 15:09:16.474 19490-19597/tileview.demo W/System.err:     at android.graphics.BitmapFactory.decodeStreamInternal(BitmapFactory.java:635)
02-16 15:09:16.474 19490-19597/tileview.demo W/System.err:     at android.graphics.BitmapFactory.decodeStream(BitmapFactory.java:611)
02-16 15:09:16.474 19490-19597/tileview.demo W/System.err:     at tileview.demo.provider.BitmapHttpProvider.getBitmap(BitmapHttpProvider.java:35)
02-16 15:09:16.474 19490-19597/tileview.demo W/System.err:     at com.qozix.tileview.tiles.Tile.generateBitmap(Tile.java:136)
02-16 15:09:16.474 19490-19597/tileview.demo W/System.err:     at com.qozix.tileview.tiles.TileCanvasViewGroup.generateTileBitmap(TileCanvasViewGroup.java:255)
02-16 15:09:16.474 19490-19597/tileview.demo W/System.err:     at com.qozix.tileview.tiles.TileRenderPoolExecutor$TileRenderRunnable.run(TileRenderPoolExecutor.java:77)
02-16 15:09:16.474 19490-19597/tileview.demo W/System.err:     at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:422)
02-16 15:09:16.474 19490-19597/tileview.demo W/System.err:     at java.util.concurrent.FutureTask.run(FutureTask.java:237)
02-16 15:09:16.474 19490-19597/tileview.demo W/System.err:     at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
02-16 15:09:16.474 19490-19597/tileview.demo W/System.err:     at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
02-16 15:09:16.474 19490-19597/tileview.demo W/System.err:     at java.lang.Thread.run(Thread.java:818)
02-16 15:09:16.474 19490-19597/tileview.demo D/skia: ---- read threw an exception
02-16 15:09:16.486 19490-19505/tileview.demo I/art: Background partial concurrent mark sweep GC freed 14805(717KB) AllocSpace objects, 201(20MB) LOS objects, 29% free, 37MB/53MB, paused 9.933ms total 52.119ms
02-16 15:09:16.537 19490-19598/tileview.demo W/System.err: java.io.InterruptedIOException
02-16 15:09:16.537 19490-19598/tileview.demo W/System.err:     at com.android.okio.Deadline.throwIfReached(Deadline.java:56)
02-16 15:09:16.537 19490-19598/tileview.demo W/System.err:     at com.android.okio.Okio$2.read(Okio.java:110)
02-16 15:09:16.537 19490-19598/tileview.demo W/System.err:     at com.android.okio.RealBufferedSource.read(RealBufferedSource.java:48)
02-16 15:09:16.537 19490-19598/tileview.demo W/System.err:     at com.android.okhttp.internal.http.HttpConnection$FixedLengthSource.read(HttpConnection.java:442)
02-16 15:09:16.537 19490-19598/tileview.demo W/System.err:     at com.android.okio.RealBufferedSource$1.read(RealBufferedSource.java:168)
02-16 15:09:16.537 19490-19598/tileview.demo W/System.err:     at android.graphics.BitmapFactory.nativeDecodeStream(Native Method)
02-16 15:09:16.537 19490-19598/tileview.demo W/System.err:     at android.graphics.BitmapFactory.decodeStreamInternal(BitmapFactory.java:635)
02-16 15:09:16.537 19490-19598/tileview.demo W/System.err:     at android.graphics.BitmapFactory.decodeStream(BitmapFactory.java:611)
02-16 15:09:16.537 19490-19598/tileview.demo W/System.err:     at tileview.demo.provider.BitmapHttpProvider.getBitmap(BitmapHttpProvider.java:35)
02-16 15:09:16.537 19490-19598/tileview.demo W/System.err:     at com.qozix.tileview.tiles.Tile.generateBitmap(Tile.java:136)
02-16 15:09:16.537 19490-19598/tileview.demo W/System.err:     at com.qozix.tileview.tiles.TileCanvasViewGroup.generateTileBitmap(TileCanvasViewGroup.java:255)
02-16 15:09:16.537 19490-19598/tileview.demo W/System.err:     at com.qozix.tileview.tiles.TileRenderPoolExecutor$TileRenderRunnable.run(TileRenderPoolExecutor.java:77)
02-16 15:09:16.537 19490-19598/tileview.demo W/System.err:     at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:422)
02-16 15:09:16.537 19490-19598/tileview.demo W/System.err:     at java.util.concurrent.FutureTask.run(FutureTask.java:237)
02-16 15:09:16.537 19490-19598/tileview.demo W/System.err:     at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
02-16 15:09:16.537 19490-19598/tileview.demo W/System.err:     at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
02-16 15:09:16.537 19490-19598/tileview.demo W/System.err:     at java.lang.Thread.run(Thread.java:818)
02-16 15:09:16.537 19490-19598/tileview.demo D/skia: ---- read threw an exception
```

also fwiw i'm running in the genymotion emulator, on a virtual Nexus 6 with 5.0

peterLaurence:
The fact that it renders from top left to bottom right is not necessarily a sign that there no parallelism.
Threads are not launched all in the same time. The first to be launched loads the top left tile, and the last one has the bottom right tile. IF the server response is always the same, and given that when a thread post runable to render its tile as soon as it has finished to download it, we observe that.

moagrius:
@peterLaurence i understand and agree.  i was pointing it out mostly because bruno's contribution to version 1, _on my particular setup_, was very obviously **not** in order - the tiles appeared in what looked like (but of course was not really) random fashion.  

that said, i know that's not necessarily the case, and it's possible that they appear to load in order: the videos he posted for comparison [Tile Loading v1 - MultiThread](https://www.dropbox.com/s/pbiroq80yz80pky/multi_threaded.mp4?dl=0), [Tile Loading v2 - SingleThread](https://www.dropbox.com/s/fmh9y5y9zsyc8g5/single_threaded.mp4?dl=0) also show that the multithreaded version is roughly in order.  

tl;dr - all good :)

peterLaurence:
@moagrius 

Yes indeed that example shows how much multi-thread speeds up!

moagrius:
yep :)

bnsantos:
@moagrius 
So, while checking the tiles not being loaded (I'm able to reproduce it) I found out another issue. If you go to any example and you press the back the `TileCanvasViewGroup.destroy()` is going to to render again all the tiles even though you're about to leave the activity.  This happens because the `TileRenderHandler` is just calling `tileCanvasViewGroup.renderTiles()` no matter what is in the message.

peterLaurence:
@bnsantos 
Hum.. on top of my head this reminds me that in `mRenderPostExecuteRunnable` we call :
```
invalidate();
requestRender();
```
So that create an infinite loop of render requests (which is not a big deal because we don't render if `DetailLevel.computeCurrentState` returns false in `beginRenderTask`). But maybe that can explain what you're seing.

moagrius:
hm is this new, since the threading PR merge?

peterLaurence:
@bnsantos 

Can you tell me in what scenario you see some tiles not being updated?
Because i already know one. And the fix is not yet integrated.

moagrius:
@peterLaurence fwiw i always get "incomplete" (white) tiles on the new http demo

bnsantos:
@peterLaurence 
The tiles are being rendered after leaving the activity. You can see the print of memory usage 
Memory usage RealMap
The first step is opening the app, second entering the Real Map Activity and the last one leaving the activity. I'm not sure if this is an issue that just appear on MultiThread or not. I'll check that later after solving the white tiles issue.

My guess about the white tiles is that the `ThreadPoolExecutor` kills those threads after 1 second idle waiting internet. This is why we can see a lot of  `InterruptedException` in the logs.  Because the `BitmapHttpProvider` prints them and `BitmapProviderAssets` doesn't. The same behavior can be achieved in `BitmapProviderAssets` if you add a `Thread.sleep(500);` then some threads will be killed but I'm not able to see where the white tile is.

@moagrius 
can you clarify my about the `TileCanvasViewGroup.onRenderTaskCancelled()` this is called whenever the whole render (all tiles) is cancelled? If so what happen when one single tile is cancelled (due to `InterruptedException`) is there a method to be called? How the loader of the image on screen is going to know that the tile bitmap was not properly rendered???

moagrius:
> can you clarify my about the TileCanvasViewGroup.onRenderTaskCancelled() this is called whenever the whole render (all tiles) is cancelled? If so what happen when one single tile is cancelled (due to InterruptedException) is there a method to be called? How the loader of the image on screen is going to know that the tile bitmap was not properly rendered???

i can only speak to how i was doing it before this most recent PR.  `onRenderTaskCancelled` is cancelled when the AysncTask is cancelled - in fact it's just a statement within the override `AsyncTask.onCancelled` method (https://github.com/moagrius/TileView/blob/1a8dcc98fa8c67dc97b70862d1a459a2b715fd9c/tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java#L392), so the exact mechanics are behind the scenes.

There was no special mechanism in place for handling a partially decoded image.  In the `doInBackground` method of the `AsyncTask`, i ran a loop that first checked if the task was cancelled - if so, stop everything - otherwise decode the tile and push the bitmap back to the UI thread with `AsyncTask.publishProgress` where again we check if the task was cancelled, and if not cancelled, the now-populated tile (with the bitmap) was passed to `tileCanvasViewGroup.addTileToCurrentTileCanvasView`.  This is all in the `TileRenderTask` definition: https://github.com/moagrius/TileView/blob/1a8dcc98fa8c67dc97b70862d1a459a2b715fd9c/tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java#L327

does that answer your question?

peterLaurence:
@bnsantos @moagrius 

I can confirm that i could reproduce the third step of memory allocation with the former version of TileView. I don't know if it's a real issue because then i paused and resume the demo app several times and the amount of memory used by the app was stable.

moagrius:
i'd suggest we separate these issues - leave the `destroy` issue for after the threading issue.  that said, if you think that this is something more easily handled together, by all means go for it.

peterLaurence:
I have something new. My repo version of Bruno when i checked his new hhtp demo was at commit 11b4baf8ffa04f0156013ae66ac08f9cbc336329. It worked very well, and my repo did not change since then. 
But now with the same app version i get no tiles (all blank screen).

moagrius:
maybe github is metering the raw image downloads?

@bnsantos do you think you'll have a working copy fairly soon, or should i revert for now?

peterLaurence:
@moagrius I wasn't clear. I was talking about the TileView demo, the real android app!

moagrius:
@peterLaurence yikes!  gtk

bnsantos:
@moagrius 
Thanks for your explanation of the `TileCanvasViewGroup.onRenderTaskCancelled()`. I was calling it if a single tile was canceled. I'm still working in the white Tiles not sure exactly what's happening but I'll need more time for sure

Now regarding the `destroy` I was not able to reproduce it in the previous version. This could be a simple demo issue or a lib issue that it is rendering tiles on destroy call. I agree that we should separate those issues and try to fix on at a time

moagrius:
@bnsantos cool, sounds good.  if after the threading issue is resolved, the destroy issue is still present, i'll create another issue and we can tackle it separately.  thanks

peterLaurence:
Found the culprit! Commenting out :
```
for ( Future f : mFutureList ) {
  if( !f.isDone() ){
    f.cancel( true );
    }
}
```
in `TileRenderPoolExecutor.cancel` solves the issue. Now, let's think about what we should have done...

peterLaurence:
Found the culprit! Commenting-out :
```
for ( Future f : mFutureList ) {
  if( !f.isDone() ){
    f.cancel( true );
  }
}
```
in `TileRenderPoolExecutor.cancel` solves the issue. Now its blazing fast!

bnsantos:
@peterLaurence 
I don't believe the culprit is on this part of the code. The "working" version I have changing the `Tile`

```java
  boolean draw( Canvas canvas ) {
    if( mBitmap != null ) {
      canvas.drawBitmap( mBitmap, mLeft, mTop, getPaint() );
      return getIsDirty();
    }else {
      return true;
    }
  }
```

Those Http requests can be cancelled in so many different situations (if they exceed the keep alive time from `ThreadPoolExecutor` which is 1s for instance) and when they're cancelled the Runnable does not know that the `BitmapProvider` failed and the `mBitmap` is null. That's why this validations on `Tile.draw`is going to force the upper layer to re-render. 

I believe that this is a fault in the architecture of the lib. Because it was planned to work on tiles that are inside the APK. If the files are not in the APK we can't assume that the `BitmapProvider` is going to be able to render all tiles.

Also there are some changes in `TileRenderPoolExecutor`because it was calling `TileCanvasViewGroup.onRenderTaskCancelled` if a single tile failed or was interrupted.

But I'm not sure that this is a proper solution or even if it is the solution. Since it is multi thread is hard to reproduce the errors in an easy repeatable pattern. I would like you to keep working on your version so we can discuss later which one achieved the best fix for this issue.

Thanks

bnsantos:
@peterLaurence 
in your test can you please add
`Thread.sleep(400);` inside your `BitmapHttpProvider` right below line 32 and just tell me the results?

peterLaurence:
@bnsantos the result : tiles are much slower to display. But they all do. No `InterruptedException`.

bnsantos:
So could be the `mFutureList` think. The whole deal of the `Future` is to be able to kind of cancel the running thread, for instance if a Tile is being downloaded and you pan or zoom then we can cancel the download and do not have to wait it download and render on screen to start rendering new tiles.

My version seems to be working, I'll keep testing it and try to have a better understanding on the `Future` think. Thanks @peterLaurence

peterLaurence:
Yes i understand that the purpose of cancelling futures is to cancel ongoing downloads, thus we avoid necessary downloads. That should be ok. But the fact disabling this functionality "solves" the issue is not good at all.
We should definitely understand what's going on.

bnsantos:
@moagrius 
Let's say a Tile bitmap was not rendered and the mBitmap is still null is this a proper solution 

```java
  boolean draw( Canvas canvas ) {
    if( mBitmap != null ) {
      canvas.drawBitmap( mBitmap, mLeft, mTop, getPaint() );
      return getIsDirty();
    }else {
      return true;
    }
  }
```

so the parent will check that the tile was not loaded and then going to put the tile on tiles to render again?

peterLaurence:
@bnsantos 
Thinking about what you've done, i will sum up here what i understand of what happens. When we pan, a bunch of tiles are requested to render. But during the pan another request cause the pending render task to be canceled and the new one gets the updated list of tiles to render. Then, all the threads that have just been cancelled may have a null bitmap. Btw, i don't undertand why because we should be protected against that. By tagging the `Tile` dirty when the bitmap is null, you use the existing mechanism that invalidates the `TileCanvasView` until there is no dirty tile left. After that, it calls `TileCanvasViewGroup.cleanup()`.
So unless i am missing something, this does not mark the tile to be rendered again.

Apart from that, i don't believe that cancelling the existing futures when a new bunch tile is requested to render is a good solution. Be cause the new tiles to render may include some for which we have already created a thread to render. We should instead make a check and and create threads if necessary.

moagrius:
> Let's say a Tile bitmap was not rendered and the mBitmap is still null is this a proper solution...  so the parent will check that the tile was not loaded and then going to put the tile on tiles to render again?

probably not going to work.  here's basically what happens (in psuedo-code) :

```
boolean isDirty = false;
for( Tile tile : tilesToBeRendered ) {
  isDirty = tile.draw( canvas );
}
if( isDirty ) {
  parentViewGroup.invalidate();
}
```

it's just checking if we need to call `invalidate` on the parent or not.

peterLaurence:
@moagrius we're on the same page :)

peterLaurence:
This time i think i got the root cause. I added this log in `TileCanvasViewGroup`:

```
void addTileToCurrentTileCanvasView( final Tile tile ) {
  if (tile.getBitmap() == null){
    System.out.println("Et voilà!!");
  }
  mTileRenderHandler.post( new PrepareTileForRenderRunnable( tile ) );
}
```
So it is possible to have a thread calling `tileCanvasViewGroup.addTileToCurrentTileCanvasView( tile )` with a null `BitMap`. Unfortunately, this cause the launch of our `PrepareTileForRenderRunnable` which in turn adds the list to the tiles that we consider to have been rendered. This is due to `mTilesAlreadyRendered.add( tile )` in `TileCanvasViewGroup.prepareTileForRender`. Subsequent request render won't request this tile to be rendered until a proper cleanup happens.

peterLaurence:
So what about :
```
void addTileToCurrentTileCanvasView( final Tile tile ) {
  if (tile.getBitmap() != null){
    mTileRenderHandler.post(new PrepareTileForRenderRunnable(tile));
  }
}
```

or in the executor :

```
if ( tile.getBitmap() != null ) {
  tileCanvasViewGroup.addTileToCurrentTileCanvasView(tile);
}
```

moagrius:
@peterLaurence i believe i'm following.  i'd say the second option seems more reasonable.

also, i see this:

```
if( !tileCanvasViewGroup.getRenderIsCancelled() && tile != null ){
  tileCanvasViewGroup.generateTileBitmap( tile );
  if ( thread.isInterrupted() ){
    tileCanvasViewGroup.onRenderTaskCancelled();
    return;
  }
  tileCanvasViewGroup.addTileToCurrentTileCanvasView( tile );
}
```

So `tileCanvasViewGroup.generateTileBitmap( tile );` might take a very long time (e.g., a network request).  We should be testing cancelled again here.  Combining that, and your second option, what about something like: 

```
if( !tileCanvasViewGroup.getRenderIsCancelled() && tile != null ){
  tileCanvasViewGroup.generateTileBitmap( tile );
  if ( thread.isInterrupted() ){
    tileCanvasViewGroup.onRenderTaskCancelled();
    return;
  }
  if( !tileCanvasViewGroup.getRenderIsCancelled() && tile.getBitmap() != null ) {
    tileCanvasViewGroup.addTileToCurrentTileCanvasView( tile );
  }
}
```

peterLaurence:
@moagrius Tested. It works :)

moagrius:
nice!  what's next?  a new PR?  @bnsantos do you want to handle this differently, e.g., i roll back, you integrate fixes, continue whatever else you're working on, and start over?  i'm open to any suggestion, both

bnsantos:
Hey Guys,

Thinking more about the solution I proposed last night I really believe it's not a nice way to do it because we can generate overhead in the parent going through all the tiles and drawing them again in the screen. I believe that in order to solve this issue we can have two options: 
1. Make parent check for null bitmaps and render only the ones that are null if they're being displayed in screen
2. Add the thread to render the tile again in the `ThreadPoolExecutor`


So I really liked the solution proposed by @peterLaurence ( which is kind 2). @moagrius regarding your comment on this part of the code
```java
if( !tileCanvasViewGroup.getRenderIsCancelled() && tile != null ){
  tileCanvasViewGroup.generateTileBitmap( tile );
  if ( thread.isInterrupted() ){
    tileCanvasViewGroup.onRenderTaskCancelled();
    return;
  }
  tileCanvasViewGroup.addTileToCurrentTileCanvasView( tile );
}
```
I need to refactor a lot of this, because it is not supposed to call `tileCanvasViewGroup.onRenderTaskCancelled()` if a single tile gets interrupted or some other exception. This method should only be called when the whole rendering (of all tiles) is cancelled.

I'll do this validation inside `ThreadPoolExecutor`
```java
if( !tileCanvasViewGroup.getRenderIsCancelled() && tile.getBitmap() != null ) {
    tileCanvasViewGroup.addTileToCurrentTileCanvasView( tile );
  }
```
And not in the `TileCanvasViewGroup`. If after testing it is ok I'll do a new PR

peterLaurence:
> I need to refactor a lot of this, because it is not supposed to call tileCanvasViewGroup.onRenderTaskCancelled() if a single tile gets interrupted or some other exception. This method should only be called when the whole rendering (of all tiles) is cancelled.

Indeed. Even if `TileView.onRenderTaskCancelled()` currently does nothing, that should be addressed. We missed that!

bnsantos:
@moagrius do you want a PR on master or in any other branch?

peterLaurence:
@bnsantos it is 7:30 AM in Texas. So unless Mike gets up early, you may just PR on master :)

moagrius:
i'm an early riser!  @bnsantos if you feel confident that your most recent update is solid, then let's go ahead and PR on master.  if you feel like we'll probably need more time, testing and changes, then PR on another branch, and let me know so I can roll back my master as well.

thank you!

bnsantos:
Ok let's do it on master!

PR #277, opened 2016-03-15
Fix #274 : double-tap change.

### LINKED ISSUE #274: Little change to double-tap
As we are waiting for the 2.1.0 version, i have another little subject to discuss.
Now, on double-tap the view is just zoomed to the actual scale times 2. And if the resulting scale exceeds the maximum, we scroll to the minimum.

We could make a little improvement based on the fact that we have better render performances at scales that are a power of 2 (as you know at those scales, tiles are rendered at their original size). So if the double-tap sets the scale to the next scale that is a power of 2, a user that double-taps and pan will have a (little but still observable) better experience. So basically if the original scale is not a power of 2, say 1.5, a double-tap will set the scale to 2. Next double-tap to 4, etc.. 
Compared to before, it increases the odds to be exactly at a scale were render performances are optimal.
Again, i have already done this and it would be easy to PR.

peterLaurence:
Fix #274 : double-tap change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Javadocs are [here](http://moagrius.github.io/TileView/index.html?com/qozix/tile
 ###Installation
 Gradle:
 ```
-compile 'com.qozix:tileview:2.1.0'
+compile 'com.qozix:tileview:2.1.1'
 ```
 
 The library is hosted on jcenter, and is not currently available from maven.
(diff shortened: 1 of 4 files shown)


peterLaurence:
I added a functionnality i really wanted but i didn't talk about : the ability to _disallow_ the return to minimum scale while being at max scale. While this is good for demo purpose, on a big map a user would disturbed if the double tap is not intentional and thus force the user to zoom in and find again his area of interest.

moagrius:
+1 looks good.  can you bump version in gradle and readme?  2.1.1 (unless you wanted to include other stuff in the next release, we can wait and bundle them, maybe a new PR for version bump)

peterLaurence:
Ok lets do that (i'm back in one hour).

PR #278, opened 2016-03-15
Fix #273 : save state.

### LINKED ISSUE #273: Restore state on orientation change
Hey Mike,

On an orientation change, and upon classic scenario when the activity recreates the TileView, it is TileView's responsibility to restore its state. Here is a good [link](http://www.intertech.com/Blog/android-custom-view-tutorial-part-3-saving-state/).
Today, TileView doesn't save the last zoom nor the position. It would be a nice default behavior not to lose the zoom and position when switching to landscape mode.
In fact, i have already done this and i'd like to know if you are interested for a PR.

peterLaurence:
Fix #273 : save state.

### DIFF
--- a/tileview/src/main/java/com/qozix/tileview/TileView.java
+++ b/tileview/src/main/java/com/qozix/tileview/TileView.java
@@ -859,9 +859,9 @@ public class TileView extends ZoomPanLayout implements
    */
   private static class SavedState extends BaseSavedState {
     /* This will store the current scale and position */
-    float scale;
-    int savedCenterX;
-    int savedCenterY;
+    float mScale;
+    int mSavedCenterX;
+    int mSavedCenterY;
 
     SavedState( Parcelable superState ) {
       super( superState );
@@ -869,17 +869,17 @@ public class TileView extends ZoomPanLayout implements
 
     private SavedState( Parcel in ) {
       super( in );
-      scale = in.readFloat();
-      savedCenterX = in.readInt();
-      savedCenterY = in.readInt();
+      mScale = in.readFloat();
+      mSavedCenterX = in.readInt();
+      mSavedCenterY = in.readInt();
     }
 
     @Override
     public void writeToParcel( Parcel out, int flags ) {
       super.writeToParcel( out, flags );
-      out.writeFloat( scale );
-      out.writeInt( savedCenterX );
-      out.writeInt( savedCenterY );
+      out.writeFloat( mScale );
+      out.writeInt( mSavedCenterX );
+      out.writeInt( mSavedCenterY );
     }
 
     public static final Parcelable.Creator<SavedState> CREATOR
@@ -905,9 +905,9 @@ public class TileView extends ZoomPanLayout implements
   public Parcelable onSaveInstanceState() {
     Parcelable superState = super.onSaveInstanceState();
     SavedState ss = new SavedState( superState );
-    ss.scale = getScale();
-    ss.savedCenterX = getScrollX() + getHalfWidth();
-    ss.savedCenterY = getScrollY() + getHalfHeight();
+    ss.mScale = getScale();
+    ss.mSavedCenterX = getScrollX() + getHalfWidth();
+    ss.mSavedCenterY = getScrollY() + getHalfHeight();
     return ss;
   }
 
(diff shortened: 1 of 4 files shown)


tileview/src/main/java/com/qozix/tileview/TileView.java
```diff
@@ -850,4 +852,81 @@ public void submit() {
...
   }
 
+  /**
+   * Object used to keep some data when a configuration change happens and the activity is
+   * re-created.
+   * It's boiler-plate but this is how to save View state.
+   */
+  private static class SavedState extends BaseSavedState {
+    /* This will store the current scale and position */
+    float scale;
```
moagrius: can we use the `m` prefix on member variables?  `mScale`, `mSavedCenterX`, `mSavedCenterY`

tileview/src/main/java/com/qozix/tileview/TileView.java
```diff
@@ -850,4 +852,81 @@ public void submit() {
...
+   * same). For this reason and if a new {@link TileView} instance is created, we have to save
+   * the current scale and position on the map, to restore them later when the {@link TileView} is
+   * recreated.
+   */
+  @Override
+  public Parcelable onSaveInstanceState() {
+    Parcelable superState = super.onSaveInstanceState();
+    SavedState ss = new SavedState( superState );
+    ss.scale = getScale();
+    ss.savedCenterX = getScrollX() + getHalfWidth();
```
moagrius: i don't think you need to save the offsets here, since you're setting the scale anyway, just save and restore scrollX/Y...

peterLaurence: In fact i think we have to save the center because on a config change, the top-left corner's position in pixel changes.

moagrius:
commented

tileview/src/main/java/com/qozix/tileview/TileView.java
peterLaurence: ok will do

tileview/src/main/java/com/qozix/tileview/TileView.java
moagrius: ah good point - you've confirmed that everything "looks right" as you have it?

tileview/src/main/java/com/qozix/tileview/TileView.java
moagrius: hm wait - these are package level, but accessed elsewhere - let me double check the style guide...

tileview/src/main/java/com/qozix/tileview/TileView.java
peterLaurence: Yes i confirm. I confirm too that saving center is needed. Just tested without it and it doesn't work.

tileview/src/main/java/com/qozix/tileview/TileView.java
moagrius: yeah package-private is `m` https://source.android.com/source/code-style.html - i guess you can just reference them directly in the same package although that looks a little weird, or add getters, up to you

tileview/src/main/java/com/qozix/tileview/TileView.java
moagrius: :thumbsup:

tileview/src/main/java/com/qozix/tileview/TileView.java
peterLaurence: Ok, i will just add `m`

tileview/src/main/java/com/qozix/tileview/TileView.java
```diff
@@ -850,4 +852,81 @@ public void submit() {
...
+    return ss;
+  }
+
+  @Override
+  public void onRestoreInstanceState(Parcelable state) {
+    final SavedState ss = (SavedState) state;
+    super.onRestoreInstanceState( ss.getSuperState() );
+    setScale( ss.scale );
+
+    /**
```
moagrius: also `/**` indicates javadoc, please use `/*` one star for a simple comment

peterLaurence: ok

peterLaurence:
Code style fixed.

moagrius:
+1 looks good, thank you

peterLaurence:
Great!

moagrius:
anything else coming or should i publish 2.1.1 to jcenter?

peterLaurence:
Nothing that come to mind right now ^^

moagrius:
it's up, thanks again sir!

PR #291, opened 2016-04-10
adding methods to convert pixel values to relative coordinates

moagrius:
Adding comments to existing methods, and 2 new methods to convert pixel values to relative coordinates.  

@peterLaurence can you do a sanity check on the 2 new methods for me?  I believe they're correct but TBH I'm doing this in a rush, and I trust your judgement.

### DIFF
--- a/tileview/src/main/java/com/qozix/tileview/geom/CoordinateTranslater.java
+++ b/tileview/src/main/java/com/qozix/tileview/geom/CoordinateTranslater.java
@@ -26,12 +26,23 @@ public class CoordinateTranslater {
 
   private boolean mHasDefinedBounds;
 
-
+  /**
+   * Set size in pixels of the image at 100% scale.
+   * @param width Width of the tiled image in pixels.
+   * @param height Height of the tiled image in pixels.
+   */
   public void setSize( int width, int height ) {
     mWidth = width;
     mHeight = height;
   }
 
+  /**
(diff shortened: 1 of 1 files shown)


peterLaurence:
Ok, i will do that soon.

moagrius:
updated.  i added convenience methods to include scale, but left off the event-specific one - i think it's reasonable for the user to supply the actual offset (scroll + event coordinate), since A) that's not a TileView specific thing, and event coordinates are always relative to the screen, and B) for a conveneince method to account for both, we'd have to have a reference to the VIew (i.e., the TileView instance) and the event, which makes for a kind of clunky API - we could stick it on TileView itself (so we get implicit reference to the scroll positions), but that seems like clutter.  I think the `translateAndScale` methods, and asking the user to supply getScrollX() + event.getX() is reasonable.  what do you think?

moagrius:
we could add something to wiki or FAQ as well...

peterLaurence:
I agree. In fact, that's what i did in the first place except that i didn't use touch event (i just retrieved the coordinates of the center of the screen). But that knowledge could be added to the wiki or the readme.md in a section like "How do i... get relative coordinates from a touch event" or something like that.

moagrius:
yep, will do.  +1 on this?

peterLaurence:
yes

PR #305, opened 2016-05-11
Fix #304 & OOM on zoom-out

### LINKED ISSUE #304: Real to relative coordinates
I am embarrassed because i just found an error in the newly added `translateAbsoluteToRelativeX` and `translateAbsoluteToRelativeY`, despite the code review.
I was astonished to get the same relative coordinates whatever the real coordinates were given to those methods. Look at this :
```
public double translateAbsoluteToRelativeY( int y ) {
  double factor = y / mHeight;
  return mTop + (factor * mDiffY);
```

`y` and `mHeight` are `int`s. And, in a normal use case (touch inside the TileView), `y` < `mHeight`. So `factor` = 0 and we always get `mTop` as returned value. Sorry i haven't noticed that Mike.

On top of that, the recent comment of @ohoedneboe should have warned us :

> Just one comment on issue #288. I see that there has been added methods
public double translateAndScaleAbsoluteToRelativeY( int y, float scale ) 
but do these functions really take the pan into effect?
I did my own implementation of this, like described in the issue thread, but I had to add the pan factor, or I got the wrong answer. My method looked like this 
public double realToRelativeY(int y, float scale, double relativeOffset)

So, @ohoedneboe, i suppose you have your own implementation of those methods, right? Because right now, the official methods can't work.

Once i fixed that issue, by just having :

```
public double translateAbsoluteToRelativeY( int y ) {
  return mTop + (y * mDiffY / mHeight);
}
```
i noticed a precision loss when getting relative coordinates from a scale different than 1 **and** when working with detail levels defined for scales > 1
For example, let's say we have a map which dimension at scale 1 is 1000px width and 1000px height.
We zoom to scale 30 so the dimensions of the map is 30000px squared.
We touch the map somewhere at X=15000px
Following the current implementation, the unscaled X is (int) (15000 / 30 + 0.5) = 500
So we see that to get a difference of 1px to the unscaled X, the delta of two touch event must be greater than (1 - 0.5) * 30 = 15px
If the second touch is at 15014px, we will get the same relative coordinates.
This particular issue won't happen for detail levels defined for scales < 1.

Fixing this was simple but required a **breaking change** :

```
public double translateAndScaleAbsoluteToRelativeY( float y, float scale ) {
  return translateAbsoluteToRelativeY( y / scale);
}
```

```
public double translateAbsoluteToRelativeY( float y ) {
  return mTop + (y * mDiffY / mHeight);
}
```

Also, when getting a `MotionEvent`, `getX` and `getY` already return `float` values, so no more cast is required.

peterLaurence:
Issue fixed, and added method signatures using `float` instead of `int`, actually only useful when some or all detail levels are defined for scale > 1.
To avoid breaking change, methods have been added and pipe to the `float` version ones.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Javadocs are [here](http://moagrius.github.io/TileView/index.html?com/qozix/tile
 ###Installation
 Gradle:
 ```
-compile 'com.qozix:tileview:2.1.5'
+compile 'com.qozix:tileview:2.1.6'
 ```
 
 The library is hosted on jcenter, and is not currently available from maven.
@@ -56,7 +56,7 @@ A demo application, built in Android Studio, is available in the `demo` folder o
 at the 2nd column from left and 3rd row from top.
 1. Create a new application with a single activity ('Main').
 1. Save the image tiles to your `assets` directory.
-1. Add `compile 'com.qozix:tileview:2.1.5'` to your gradle dependencies.
+1. Add `compile 'com.qozix:tileview:2.1.6'` to your gradle dependencies.
 1. In the Main Activity, use this for `onCreate`:
 ```
 @Override
--- a/tileview/build.gradle
+++ b/tileview/build.gradle
@@ -6,8 +6,8 @@ android {
     defaultConfig {
         minSdkVersion 11
         targetSdkVersion 22
-        versionCode 30
-        versionName "2.1.5"
+        versionCode 31
+        versionName "2.1.6"
     }
     buildTypes {
         release {
--- a/tileview/src/main/java/com/qozix/tileview/TileView.java
+++ b/tileview/src/main/java/com/qozix/tileview/TileView.java
@@ -85,6 +85,7 @@ public class TileView extends ZoomPanLayout implements
   private RenderThrottleHandler mRenderThrottleHandler;
 
   private boolean mShouldRenderWhilePanning = false;
+  private boolean mShouldUpdateDetailLevelWhileZooming = false;
 
   /**
    * Constructor to use when creating a TileView from code.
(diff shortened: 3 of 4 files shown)


peterLaurence:
Also, after thoughts, the use of `FloatMathHelper` makes no sense in that context as we want to get relative coordinates (so `doubles`s) and there is no reason to have an unscaledX/Y being an `int`.
Again, the benefit of doing that is virtually null when working with 0 < scale < 1, but is very important when detail levels are defined for scale > 1.

moagrius:
LGTM +1, thanks Peter

moagrius:
i can merge, version bump, and generate artifacts, or you can (at least the first 2), up to you, let me know

peterLaurence:
I can but only tomorrow evening (in about 20 hours).

moagrius:
either way.  if you get to it, great, otherwise I will when I can.  thx

peterLaurence:
We can also include my next PR for the OOM fix. I think that i will be able to do that after work.
It would save you one artifact generation.

moagrius:
:thumbsup:

moagrius:
feel free to merge whenever you want (or i can), or we can wait for your other PR...  just LMK when to generate a release

peterLaurence:
My bad, i did not create another PR. I just added the commit "fix OOM on zoom-out". I can rename the branch so it isn't called "fix #304" (and the title of this PR).

peterLaurence:
I just updated the name of the PR.

moagrius:
looks great, +1

moagrius:
aar generated and on bintray, good to go, thanks Peter

PR #325, opened 2016-08-07
using a single canvas

moagrius:
This release changes how tiles are rendered.  Instead of a `View` for each detail level, drawn in order, we use a single `ViewGroup` `canvas` to draw all tiles.

Some APIs have been deprecated, most notably `shouldRecycleBitmaps` - bitmaps are always recycled now.

Other changes should be largely internal and not noticed by most users, and should generally considered to be non-breaking, however some changes could be noticed (for example, `TileCanvasViewGroup` no longer has `TileCanvasView` children - or any children - so if an existing app uses `TileView.getTileCanvasViewGroup().getChildAt(0)`, the result would now be null.

Changes that could be considered breaking:
- `TileCanvasView` has been removed
- `TileCanvasViewGroup` no longer extends `ScalingLayout`, or implements `TileCanvasView.TileCanvasDrawListener`
- `Tile.stampTime` is deprecated and now a no-op


moagrius:
@peterLaurence ready for PR.  there was actually a ton left to do, but i think i've found everything, squashed all the bugs, and supported all the features and behaviors we provide.  I've run it on real devices and emus, using the labs demo and the primary repo (this) demos, but it could always use some more testing if you have the time, on whatever devices or emus you have available and time to try.  post back with any questions or suggestions.  thanks

peterLaurence:
@moagrius ok, i will try this and report back.

tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
```diff
@@ -143,103 +175,207 @@ public void suppressRender() {
...
+   *
+   * @param canvas The Canvas instance to draw tile bitmaps into.
+   */
+  private void drawTiles( Canvas canvas ) {
+    if( mPreviouslyDrawnTiles.size() > 0 ) {
+      drawTilesConsideringPreviouslyDrawnLevel( canvas );
+    } else {
+      drawTilesWithoutConsideringPreviouslyDrawnLevel( canvas );
     }
+    Log.d( getClass().getSimpleName(), "prevous tile count: " + mPreviouslyDrawnTiles.size() );
```
peterLaurence: TODO : remove that log before release.

peterLaurence:
So i have tested this branch with the demos and my personal app. The behavior is very good, no issue so far. One particular point is that the tile transitions doesn't affect the overall performance and even improve the user experience (that's my opinion). Perhaps we should update the demos, for example `LargeImageTileViewActivity`, where we can see :

```
// by disabling transitions, we won't see a flicker of background color when moving between tile sets
tileView.setTransitionsEnabled( false );
```

Also, i noticed that the demo gradle has `targetSdkVersion 23`, whereas the TileView gradle `targetSdkVersion 22`. Is there a reason for this?

moagrius:
sounds great.  I agree on updating the demos, and I'll fix the target SDKs.  thanks Peter, +1?

peterLaurence:
Yes, +1 (with the debug log removed).

moagrius:
:thumbsup:

PR #358, opened 2016-10-21
added rubber band like scrolling effect

krabo0om:
I've added a simple setter to allow scrolling outside of the image. A (nice) side effect is a rubber band like effect, which should help to visualize the edge of the image.
It works most of the times, but if it's a drag and not a fling the view will not bounce back. Although we could sell this as a feature :D
However, I didn't had time to look into it further, but if you like this feature and don't see a breaking change I'll finalize it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Javadocs are [here](http://moagrius.github.io/TileView/index.html?com/qozix/tile
 ###Installation
 Gradle:
 ```
-compile 'com.qozix:tileview:'2.2.4'
+compile 'com.qozix:tileview:'2.2.5'
 ```
 
 The library is hosted on jcenter, and is not currently available from maven.
@@ -64,7 +64,7 @@ A demo application, built in Android Studio, is available in the `demo` folder o
 at the 2nd column from left and 3rd row from top.
 1. Create a new application with a single activity ('Main').
 1. Save the image tiles to your `assets` directory.
-1. Add `compile 'com.qozix:tileview:2.2.4'` to your gradle dependencies.
+1. Add `compile 'com.qozix:tileview:2.2.5'` to your gradle dependencies.
 1. In the Main Activity, use this for `onCreate`:
 ```
 @Override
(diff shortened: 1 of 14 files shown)


moagrius:
very cool.  i'm not sure how a fling would bounce back (sounds like it's shows interpolation from the word "bounce" with just those offsets, but i haven't had a chance to run it (i'll try to do so this weekend).

i'm not sure this feature would be used by enough people to add it to the core api, but i am in the first stages of version 3, which is going to use a plugin system for things like paths, hotspots, markers, etc - effects like this might also be a good fit.

krabo0om:
I reworked and polished it. I replaced the Scroller with an OverScroller and adjusted the getConstrainedScroll() methods.

moagrius:
does this have any effect whatsoever if the option is _not_ explicitly turned on?  will existing apps with TileView notice *any difference* if this change is incorporated?

MisterRager:
@krabo0om you should be able to implement this off the trunk w/o rewriting any base code, now

krabo0om:
@MisterRager hey, sorry it took me so long to respond. I checked out your changes from 8ff22db92fe08be53fddb55e20bb27c92431c936 and overwrote `getScrollMinX()` but this solution has a lot of drawbacks compared to mine.
First, it does start with the offset, so the content is not centered.
Second, no smooth scrolling, hard snaps.
Third, it does not work for every side. I.e. I can scroll further up, but not further down.
I'll check out #384 next and compare it.

krabo0om:
I also have to apologize to you @moagrius for the months of silence.
The demos work fine with the changed scrolling effects (meaning they work like before). Checkout the fictional map from my [master](https://github.com/krabo0om/TileView), a tap anywhere will toggle the effect.

moagrius:
np.  let's see how this other feature set turns out.  i'll leave the PR open

krabo0om:
Feel free to close this pull request, for now I will just use the simple additional whitespace feature from #384.
Maybe someone will reimplement it in the new version 3.

moagrius:
:thumbsup:

PR #384, opened 2017-01-06
Scroll changes - allow fling to slide off left and top; allow scroller to be replaced

MisterRager:
This changes the logic in `ZoomPanLayout#onFling` to use `#getScrollMinX` and `#getScrollMinY` to decide how far to the left and top it will scroll instead of using hard-coded `0`.

### DIFF
--- a/tileview/src/main/java/com/qozix/tileview/widgets/ZoomPanLayout.java
+++ b/tileview/src/main/java/com/qozix/tileview/widgets/ZoomPanLayout.java
@@ -713,7 +713,9 @@ public class ZoomPanLayout extends ViewGroup implements
 
   @Override
   public boolean onFling( MotionEvent event1, MotionEvent event2, float velocityX, float velocityY ) {
-    mScroller.fling( getScrollX(), getScrollY(), (int) -velocityX, (int) -velocityY, 0, getScrollLimitX(), 0, getScrollLimitY() );
+    mScroller.fling( getScrollX(), getScrollY(), (int) -velocityX, (int) -velocityY,
+                     getScrollMinX(), getScrollLimitX(), getScrollMinY(), getScrollLimitY() );
+
     mIsFlinging = true;
     ViewCompat.postInvalidateOnAnimation( this );
     broadcastFlingBegin();
--- a/tileview/src/main/java/com/qozix/tileview/widgets/ZoomPanLayout.java
+++ b/tileview/src/main/java/com/qozix/tileview/widgets/ZoomPanLayout.java
@@ -589,18 +589,18 @@ public class ZoomPanLayout extends ViewGroup implements
 
   @Override
   public void computeScroll() {
-    if( mScroller.computeScrollOffset() ) {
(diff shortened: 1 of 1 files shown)


moagrius:
LGTM.  @peterLaurence ?

peterLaurence:
@moagrius i will take the time to look at this tomorrow morning.

peterLaurence:
i'm ok with the changes in the last commit, but i don't understand why we should instantiate the `Scroller` in the getter, while there's no setter for it. In other words, in the main branch, there is no way to change the `Scroller` and internally referring to the private field makes sense.
In the proposed changes there is no way either to change the `Scroller`.
@MisterRager  can you explain why you did this?

peterLaurence:
Well i have responded to myself..  to change the `Scroller`, one can override the `getScroller` method. 
It's just that i would have expected a setter for that (which means new api). I fell like forcing the user to extend `ZoomPanLayout` just to change a `Scroller` is a bit too much.

krabo0om:
I appreciate the changes, but why only top and left and not down and right?
To get support for all four directions the user would have to override not only `getScrollMinX()` and `getScrollMinY()` but also `getScrollLimitX()` and `getScrollLimitY()`. I, as a user, would like a single method to override and be done with all 4 cases.
Also, when using `MinimumScaleMode.FIT` and `setScale(0f)` the content is not centred, but shifted to the bottom right.

tileview/src/main/java/com/qozix/tileview/widgets/ZoomPanLayout.java
```diff
@@ -713,7 +716,9 @@ public boolean onDown( MotionEvent event ) {
 
   @Override
   public boolean onFling( MotionEvent event1, MotionEvent event2, float velocityX, float velocityY ) {
-    mScroller.fling( getScrollX(), getScrollY(), (int) -velocityX, (int) -velocityY, 0, getScrollLimitX(), 0, getScrollLimitY() );
+    getScroller().fling( getScrollX(), getScrollY(), (int) -velocityX, (int) -velocityY,
```
MisterRager: @krabo0om this is why I only added the top and left: I had forgotten them before.

MisterRager:
@peterLaurence that's fair. Should I add a setter, as well?

peterLaurence:
@MisterRager if you add a setter, then the default `Scroller` instance should be set in the constructor. The getter would just return the current instance. Then, the new `setScroller` would be a new api, and this should be reflected by a version bump. @moagrius what do you think?
If you don't change anything, well.. i won't be angry :)

moagrius:
yeah i'm fine either way :)

MisterRager:
I'd figured that since the scroll bounds are overridden by extension rather than setters, it isn't that much more painful to override scroll behavior by the same.

peterLaurence:
Ok then.

krabo0om:
@MisterRager  what is the purpose of `getScroller()`?
According to the [Android Documentation](https://developer.android.com/reference/android/widget/Scroller.html) there is no subclass.

PR #412, opened 2017-05-08
remove the `-target 8` line from proguard-rules.pro

MisterRager:
This should fix builds for you, @moagrius , but I can't confirm that this produces a library version that works with the desugar tool. 

I've got a [dockerfile](https://gist.github.com/MisterRager/e2af2178b8f6171aab85af0f4e2e19b5) that I cobbled together to try to run builds. I run the following command:
```
docker run -v "$(pwd):/werk" -w '/werk' -e JAVA_HOME=/usr/lib/jvm/java-7-openjdk-amd64 16b601585a4b  /werk/gradlew tasks
```
But I get stuck with this error:
```
> java.lang.UnsupportedClassVersionError: com/android/build/gradle/AppPlugin : Unsupported major.minor version 52.0
```

It's also crashing with someone else's docker setup with the same error (same error):
```
docker run --tty --interactive --volume=$(pwd):/opt/workspace --workdir=/opt/workspace --rm jacekmarchwicki/android:java7  /bin/sh -c "./gradlew build"
```

If you could throw me the aar or zip release produced, I can walk through the contents with `javap` to make sure stackmap frames are present.

### DIFF
--- a/tileview/proguard-rules.pro
+++ b/tileview/proguard-rules.pro
@@ -18,5 +18,3 @@
 -keep class com.qozix** {*;}
 -keepattributes InnerClasses
 -keepattributes EnclosingMethod
-
--target 8


moagrius:
we've run into that `major.minor 52` bug at work, i'll dig up the fix

moagrius:
@MisterRager upgrade AS to get rid of that `Unsupported major.minor version 52.0` bug.

should i roll back that the commits that got into master for now?

MisterRager:
@moagrius does this branch not build for you? I think that the default proguard file you had before might still end up stripping stackmap frames.

moagrius:
didn't know it was ready to try, thought you were still trying to resolve errors

i'll try it tonight

MisterRager:
Sorry for not being clear: I am certain that this branch will remove that bug from your build, and I am uncertain but fairly sure that it will return the stackmap frames to the resulting library bundle as required by the bytecode verifiers in Desugar and such.

moagrius:
this works for me, i've also uploaded the release to bintray as 2.2.6

@MisterRager can you confirm it resolved the issues you were uncertain about - sorry i'm still only about 20% checked in to what's happening with this branch - we're trying to release the product i've spent the last 10 months on at work and things are nuts

@peterLaurence are you following this?  can you confirm all's well with the repo at this point?  also, i'll be ready to get started on v3 in about 3 weeks, when summer term starts.  first thing i'd like to do is the base component - a two-dimensional scroller.  ZoomPanLayout will extend that as  it's own component, for things like zoomable images or grids.  My hope is that each component can exist separately on top of each other, until we finally get to pluggable things like paths, markers, hotspots, etc.  anything in particular you'd like to achieve?  we should probably move this out of issues, or at least out of this issue - how do you want to coordinate?  wiki, or feature issues, or something else?

MisterRager:
The lib uploaded to bintray contains the appropriate `StackMapTable` entries and is bytecode version `51`. It sounds like it's all good!

moagrius:
great, thanks @MisterRager - promise i'll pay more attention next time!

MisterRager:
I think the errors I was having in my Docker-based builds was that I was using too new an Android SDK, and the actual Android SDK version installed in the image needed to use JDK 8. I probably could have gotten it working by walking backwards in versions to find a compatible one, but that's just conjecture. Thanks for the help here. I've definitely learned way more about the JVM than I had realized I would in the course of this.

moagrius:
:thumbsup:

Issue #415, opened 2017-05-20
Offset with custom scrollMin()


krabo0om:
#384 introduced a new bug.
When overwriting `getScrollMinY`, using `setMinimumScaleMode(ZoomPanLayout.MinimumScaleMode.FIT)` and setting the scale to 0 `setScale(0.0f)` the content is shifted down or upwards with negative or positive value, respectively.
I illustrated it with a modified demo in my [showcase#415](https://github.com/krabo0om/TileView/tree/showcase%23415) branch

peterLaurence:
I tested your code. I don't see anything wrong : i see the scroll area extended by a fixed amount of pixels. Can you add a screenshot or add details?

krabo0om:
the distance between top-plan and plan-bottom is different, it is not centered vertically.
![offset_buildingplan_modded](https://cloud.githubusercontent.com/assets/3706149/26277719/eb33d6ce-3d8d-11e7-8c75-fcbe51d2b87d.png)

peterLaurence:
Yes i can see that. Will try to figure out what is happening.

peterLaurence:
My understanding so far is that the constrained scrollY logic isn't appropriate when the minimum scrollY is negative.
I will try to finish the job this week.

peterLaurence:
@krabo0om I found a fix. Its not perfect, but from my tests it's definitely better than before. You can test that on my [fix415](https://github.com/peterLaurence/TileView/tree/fix%23415) branch. You will have to add your custom `ScrollTileView`, as it can't be part of the library.

krabo0om:
works great!
thank you very much.

peterLaurence:
Now i need to make a PR, so it goes into the next release.

PR #417, opened 2017-05-24
Fix#415

peterLaurence:
Hey Mike,

I don't know if you're aware of #415 : the OP describes the problem pretty well. To summarize, when the MinimumScaleMode#FIT was introduced, the logic to calculate the constrained X/Y scroll position needed to be adapted. Well, not exactly. In fact, in most cases there is no issue. But when a developer extends the `TileView` and overrides (as he's allowed to) the `getScrollMinX` and `getScrollLimitX` to return something different than 0, then the whole TileView is shifted at minimum scale. 

Investigating on this, i found that it was not the only issue. Scrolling slowly at the boundaries of the TileView would make the scroll position change from eg -100 to 0 without any transition (which is a weird behavior). This happens when the scale is equal to `mMinimumScaleY` or `mMinimumScaleX`, depending on the shape of the tileset.

So i worked on the logic of `getConstrainedScrollX` and `getConstrainedScrollY` to work also when non zero return values are defined for `getScrollMinX`, `getScrollLimitX`, `getScrollMinY`, `getScrollLimitY`.
I came up with this implementation, which i tested and so did the author of #415.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Javadocs are [here](http://moagrius.github.io/TileView/index.html?com/qozix/tile
 ### Installation
 Gradle:
 ```
-compile 'com.qozix:tileview:'2.2.6'
+compile 'com.qozix:tileview:'2.2.7'
 ```
 
 The library is hosted on jcenter, and is not currently available from maven.
@@ -64,7 +64,7 @@ A demo application, built in Android Studio, is available in the `demo` folder o
 at the 2nd column from left and 3rd row from top.
 1. Create a new application with a single activity ('Main').
 1. Save the image tiles to your `assets` directory.
-1. Add `compile 'com.qozix:tileview:2.2.6'` to your gradle dependencies.
+1. Add `compile 'com.qozix:tileview:2.2.7'` to your gradle dependencies.
 1. In the Main Activity, use this for `onCreate`:
 ```
 @Override
--- a/tileview/build.gradle
+++ b/tileview/build.gradle
@@ -6,8 +6,8 @@ android {
     defaultConfig {
         minSdkVersion 11
         targetSdkVersion 23
-        versionCode 37
-        versionName "2.2.6"
+        versionCode 38
+        versionName "2.2.7"
     }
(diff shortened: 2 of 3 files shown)


moagrius:
+1 can you bump version, update readme and merge?  i'll pull and upload to jcenter asap

moagrius:
@peterLaurence i posted this in another PR a while back, ICYMI:

> @peterLaurence ~are you following this? can you confirm all's well with the repo at this point?~ also, i'll be ready to get started on v3 in about 3 weeks, when summer term starts. first thing i'd like to do is the base component - a two-dimensional scroller. ZoomPanLayout will extend that as it's own component, for things like zoomable images or grids. My hope is that each component can exist separately on top of each other, until we finally get to pluggable things like paths, markers, hotspots, etc. anything in particular you'd like to achieve? we should probably move this out of issues, or at least out of this issue - how do you want to coordinate? wiki, or feature issues, or something else?

peterLaurence:
I don't know how i missed your message above.. sorry for that.
Ok for your strategy to decouple first. To coordinate, i find the "Projects" tab excellent. You define a new project, say "v3", then you can add columns e.g ("To Do", "In Progress", "Done"). Each columns can receive cards describing the task. Each card can be converted into an issue, and moved to other columns. This gives an overview of the project, while still using issues for details.

As for things i'd like to achieve : have a mode that loads tiles so that the user don't even see the loading. But something efficient (like don't just extend the viewport). I don't know if its feasible. We will see :)

moagrius:
Sounds great.  I created a project, will start to fill it in tonight.  I'll also publish the latest to bintray tonight.  Also, I like your idea - we might not be able to get it perfect, but I get we can come close.  I'll ping you later.  Thanks @peterLaurence

peterLaurence:
Ok great! I will look into that tomorrow when I get time.
The timing is perfect for me, I have finished some personal dev work so I have some spare time.

moagrius:
sounds great

also, 2.2.7 is on jcenter

thanks for the fix

PR #450, opened 2017-11-08
Fix #301

### LINKED ISSUE #301: Out of memory, zooming, ZoomPanListener
Thanks for this wonderful library.  I have put together several mapping apps which display tiles downloaded from servers.  They have a fairly large number of zoom levels (up to 20).  I have also created a couple of buttons for "zoom in" and "zoom out" which double and halve the scale respectively. 
I have noticed that if I set the initial zoom level (scale) so that the map is "zoomed out", the library crashes with an out of memory error.  The exact zoom level at which this happens depends on the device.  I have worked around this by initializing the library so that the map starts out quite "zoomed in".
I have also noticed that if I start the map "zoomed in", and zoom out using my zoom buttons, no out of memory error occurs.
Also, the "zoom wraparound", in which the scale of the map goes from its minimum to maximum value when the map is double tapped also causes an out of memory error in the library.  I have intercepted this and blocked it to prevent the crash.
Also, I notice that occasionally when the zoom out button is tapped, my "getBitmap" callback is not called by the library.  If I then pan the map, my "getBitmap" callback is called and the map fills correctly.
Also, I have noticed that the current detail level is not set correctly when the "onZoomEnd" callback is called in the ZoomPanListener .
Finally, I have noticed from analyzing crashes, and just wanted to confirm, that the "getBitmap" callback is called from an AsyncTask, so there is no need for me to start up my own AsyncTask in getBitmap() when downloading a tile.
Again, thanks for the library.

peterLaurence:
This fixes the missing tiles issue i could reproduce.

There is only one change : modifying the fast-fail test in `TileCanvasViewGroup.requestRender()` and removing all boilerplate code related to `mLastRenderedDetailLevel`.

The explanation of this change is, from the tests i run,  when we have missing tiles after a quick zooming out, we can observe the following :

A high amount of request render are kicked in `TileCanvasViewGroup.requestRender()` which is based on a throttle logic. Those requests which usually pass through this enable the launch of new tile decoding with the `TileRenderPoolExecutor`, in the `beginRenderTask`. So since no new task are sent to the executor, the rendering mechanism is broken.
The addition of the `mTilesInCurrentViewport.size() > 0` in the fast-fail test enables the last request render to not fast-fail and trigger the missing render.

As for the removing of the `mLastRenderedDetailLevel`, there are two reasons :
* it was historically added before we removed the detail level lock, to fix an OOM issue. It is not needed anymore.
* it's ugly

### DIFF
--- a/tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
+++ b/tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
@@ -362,8 +362,11 @@ public class TileCanvasViewGroup extends ViewGroup {
   }
 
   private void beginRenderTask() {
-    // update the detail level properties considering the visible viewport
-    mDetailLevelToRender.computeCurrentState();
+    // if visible columns and rows are same as previously computed, fast-fail
+    boolean changed = mDetailLevelToRender.computeCurrentState();
+    if( !changed && mTilesInCurrentViewport.size() > 0 ) {
+      return;
+    }
     // determine tiles are mathematically within the current viewport; force re-computation
     mDetailLevelToRender.computeVisibleTilesFromViewport();
     // get rid of anything outside, use previously computed intersections
--- a/tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
+++ b/tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
@@ -39,8 +39,6 @@ public class TileCanvasViewGroup extends ViewGroup {
   private BitmapRecycler mBitmapRecycler;
 
   private DetailLevel mDetailLevelToRender;
-  private DetailLevel mLastRenderedDetailLevel;
-
 
   private boolean mRenderIsCancelled = false;
   private boolean mRenderIsSuppressed = false;
@@ -364,11 +362,8 @@ public class TileCanvasViewGroup extends ViewGroup {
   }
 
   private void beginRenderTask() {
-    // if visible columns and rows are same as previously computed, fast-fail
-    boolean changed = mDetailLevelToRender.computeCurrentState();
-    if( !changed && mDetailLevelToRender.equals( mLastRenderedDetailLevel ) ) {
-      return;
-    }
+    // update the detail level properties considering the visible viewport
+    mDetailLevelToRender.computeCurrentState();
     // determine tiles are mathematically within the current viewport; force re-computation
     mDetailLevelToRender.computeVisibleTilesFromViewport();
     // get rid of anything outside, use previously computed intersections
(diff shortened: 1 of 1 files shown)


tileview/src/main/java/com/qozix/tileview/tiles/TileCanvasViewGroup.java
```diff
@@ -366,7 +364,7 @@ private void markTilesAsPrevious() {
   private void beginRenderTask() {
     // if visible columns and rows are same as previously computed, fast-fail
     boolean changed = mDetailLevelToRender.computeCurrentState();
-    if( !changed && mDetailLevelToRender.equals( mLastRenderedDetailLevel ) ) {
+    if( !changed && mTilesInCurrentViewport.size() > 0 ) {
```
moagrius: is this logic right?  if there is no change, and there are no tiles, fast fail...?  shouldn't that be "or" rather than "and"?

peterLaurence: Sorry for the late answer. 
The logic says : if there are no tiles, do not fast fail.
This is sadly the only fix i found. Like explained, i had to find a way to restrain the fast fail occurrences.

moagrius: hm, i can't say i get it at this point, but i trust you, and we're pretty much in maintenance mode at this point, so i'm going to merge and push this out to jcenter.

thanks @peterLaurence!

moagrius:
@peterLaurence this is in 2.2.8, which is now live on bintray - maybe it's just me, but does this seem to affect performance for you at all?  seems a bit more sticky/jumpy, but then again i'm not often on this machine and haven't run the test app for a while...

bb-juliogarcia:
Thanks @peterLaurence & @moagrius for taking the time to look into this. I pulled this version into our app, and the blank tiles issue seems to be gone. However, I ran it on a Nexus 5X and also noticed the stickiness and jumpiness that @moagrius observed. @peterLaurence is this happening for you?

peterLaurence:
@moagrius @bb-juliogarcia This is strange because i dont have this issue on my side (even on emulator). Although i just made a fresh install on my old Nexus 5, and the first time i started the sample activity it was sticky/bumpy. Then i stopped the activity and restarted, and it was running smoothly.. strange.
I will try to see if its due to my change, or if it only happens after a fresh install of the app (which could be the case since the java runtime can "learn" from previous executions).

bb-juliogarcia:
Same thing happened for me actually, the performance is fine now. Thanks for your work! @peterLaurence

peterLaurence:
Maybe it's worth noticing that the new android profiler in Android Studio 3.0.1 affects the performance of the app, which is not a big surprise since it goes much deeper into execution analysis.

moagrius:
interesting, and gtk

Issue #463, opened 2018-01-22
Scroll outside of bounds


Derpalus:
Hello again, more questions!

I'd like to be able to scroll past the bounds of the image. Like if I'm zoomed in so the image is covering the entire screen, there is no way to scroll past the edge of the image.

I can set up so I can zoom out past the edges of the image using `setMinimumScaleMode` and `setScaleLimits`, but I still haven't figured out how (if even possible) to override the scrolling?

Derpalus:
I'm not quite sure I understand what you mean I should do?
Should I fork it? In that case it seems like I should perhaps change the values that getScrollMinX/Y and getScrollLimitX/Y provide.

moagrius:
you can subclass it and override those methods to return different values.  e.g., instead of a minimum of 0, you could have a minimum of -1000

Derpalus:
Thanks for the help!

moagrius:
NP, let us know how it goes

tkrz:
I have to do the same thing, subclassing is not a problem, but how i should supply modified class to TileView?

Derpalus:
I have made a pull request of my solution. Don't know if it's good enough but if mogarius accepts it then you can use it as well without having to make your own changes.

moagrius:
@tkrz As @Derpalus mentioned, he's submitted a PR, which I suspect will be accepted in a few days, but as to your question:

> I have to do the same thing, subclassing is not a problem, but how i should supply modified class to TileView?

You make a subclass like `MyTileView`, then use that instead of a normal `TileView`.

moagrius:
the latest release offers this feature

Issue #460, opened 2018-01-15
Absolute anchor positions for markers


Derpalus:
Hello,

I wonder if there is a way to use absolue anchor positions for markers?

My problem is that I have a marker which is a nine-patch so it changes size. And this works fine with the relative positions as long as the nine-patch changes size symmetrically, but it just so happens that mine doesn't. In my case the ancor point lies 32px in from the left side of the marker, but the marker only changes size in the right half.

Currently I'm solving this by:
```
LinearLayout marker = new ...;
Point size = new Point();
getWindowManager().getDefaultDisplay().getSize(size);
marker.measure(size.x, size.y);
int width = layout.getMeasuredWidth();
// anchorX = -32.0f / width
```

but this does not give exactly correct sizes (probably because it's a nine-patch that hasn't been inflated correctly).

moagrius:
this is in the latest release

PR #466, opened 2018-01-26
Borders and markers improvement

Derpalus:
In relation to issue #460 and #463.

I have never done any pull requests on GitHub before so I don't know if any more info is required.

### DIFF
--- a/tileview/src/main/java/com/qozix/tileview/TileView.java
+++ b/tileview/src/main/java/com/qozix/tileview/TileView.java
@@ -511,18 +511,18 @@ public class TileView extends ZoomPanLayout implements
    * @param view    View instance to be added to the TileView.
    * @param x       Relative x position the View instance should be positioned at.
    * @param y       Relative y position the View instance should be positioned at.
-   * @param anchorX The x-axis position of a marker will be offset by a number equal to the width of the marker multiplied by this value.
-   * @param anchorY The y-axis position of a marker will be offset by a number equal to the height of the marker multiplied by this value.
-   * @param typeAnchorX The x-axis position type of a marker. See POS_* flags in MarkerLayout.
-   * @param typeAnchorY The y-axis position type of a marker. See POS_* flags in MarkerLayout.
+   * @param relativeAnchorX The x-axis position of a marker will be offset by a number equal to the width of the marker multiplied by this value.
+   * @param relativeAnchorY The y-axis position of a marker will be offset by a number equal to the height of the marker multiplied by this value.
+   * @param absoluteAnchorX The x-axis position of a marker will be offset by this value.
+   * @param absoluteAnchorY The y-axis position of a marker will be offset by this value.
    * @return The View instance added to the TileView.
    */
-  public View addMarker( View view, double x, double y, Float anchorX, Float anchorY, int typeAnchorX, int typeAnchorY ) {
+  public View addMarker( View view, double x, double y, Float relativeAnchorX, Float relativeAnchorY, float absoluteAnchorX, float absoluteAnchorY ) {
     return mMarkerLayout.addMarker( view,
             mCoordinateTranslater.translateX( x ),
             mCoordinateTranslater.translateY( y ),
-            anchorX, anchorY,
-            typeAnchorX, typeAnchorY
+            relativeAnchorX, relativeAnchorY,
+            absoluteAnchorX, absoluteAnchorY
     );
   }
 
--- a/tileview/src/main/java/com/qozix/tileview/TileView.java
+++ b/tileview/src/main/java/com/qozix/tileview/TileView.java
@@ -505,6 +505,28 @@ public class TileView extends ZoomPanLayout implements
   }
 
   /**
+   * Add a marker to the the TileView.  The marker can be any View.
+   * No LayoutParams are required; the View will be laid out using WRAP_CONTENT for both width and height, and positioned based on the parameters.
+   *
+   * @param view    View instance to be added to the TileView.
+   * @param x       Relative x position the View instance should be positioned at.
+   * @param y       Relative y position the View instance should be positioned at.
+   * @param anchorX The x-axis position of a marker will be offset by a number equal to the width of the marker multiplied by this value.
+   * @param anchorY The y-axis position of a marker will be offset by a number equal to the height of the marker multiplied by this value.
+   * @param typeAnchorX The x-axis position type of a marker. See POS_* flags in MarkerLayout.
+   * @param typeAnchorY The y-axis position type of a marker. See POS_* flags in MarkerLayout.
+   * @return The View instance added to the TileView.
+   */
+  public View addMarker( View view, double x, double y, Float anchorX, Float anchorY, int typeAnchorX, int typeAnchorY ) {
+    return mMarkerLayout.addMarker( view,
+            mCoordinateTranslater.translateX( x ),
+            mCoordinateTranslater.translateY( y ),
+            anchorX, anchorY,
+            typeAnchorX, typeAnchorY
+    );
+  }
+
+  /**
    * Removes a marker View from the TileView's view tree.
    *
    * @param view The marker View to be removed.
--- a/tileview/src/main/java/com/qozix/tileview/markers/MarkerLayout.java
+++ b/tileview/src/main/java/com/qozix/tileview/markers/MarkerLayout.java
@@ -16,10 +16,6 @@ public class MarkerLayout extends ViewGroup {
 
   private MarkerTapListener mMarkerTapListener;
 
-  public static final int ANCHOR_BEHAVIOUR_RELATIVE = 0;               // Position is relative
-  public static final int ANCHOR_BEHAVIOUR_ABSOLUTE_LEFT_TOP = 1;      // Position is absolute calculated from left or top of marker
-  public static final int ANCHOR_BEHAVIOUR_ABSOLUTE_RIGHT_BOTTOM = 2;  // Position is absolute calculated from right or bottom of marker
-
   public MarkerLayout( Context context ) {
     super( context );
     setClipChildren( false );
--- a/tileview/src/main/java/com/qozix/tileview/markers/MarkerLayout.java
+++ b/tileview/src/main/java/com/qozix/tileview/markers/MarkerLayout.java
@@ -16,6 +16,10 @@ public class MarkerLayout extends ViewGroup {
 
   private MarkerTapListener mMarkerTapListener;
 
+  public static final int POS_RELATIVE = 0;               // Position is relative
+  public static final int POS_ABSOLUTE_LEFT_TOP = 1;      // Position is absolute calculated from left or top of marker
+  public static final int POS_ABSOLUTE_RIGHT_BOTTOM = 2;  // Position is absolute calculated from right or bottom of marker
+
   public MarkerLayout( Context context ) {
     super( context );
     setClipChildren( false );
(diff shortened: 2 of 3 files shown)


moagrius:
At first glance this looks pretty good - there's a couple nits about naming and method signature but I think it'll probably be very minor changes.  I need to run it a couple times and am slammed right now, let's shoot for mid week for a review.

Thanks for this.

moagrius:
i haven't forgotten about this - just no time yet - sorry

tileview/src/main/java/com/qozix/tileview/TileView.java
```diff
@@ -504,6 +504,28 @@ public View addMarker( View view, double x, double y, Float anchorX, Float ancho
...
+  /**
+   * Add a marker to the the TileView.  The marker can be any View.
+   * No LayoutParams are required; the View will be laid out using WRAP_CONTENT for both width and height, and positioned based on the parameters.
+   *
+   * @param view    View instance to be added to the TileView.
+   * @param x       Relative x position the View instance should be positioned at.
+   * @param y       Relative y position the View instance should be positioned at.
+   * @param anchorX The x-axis position of a marker will be offset by a number equal to the width of the marker multiplied by this value.
+   * @param anchorY The y-axis position of a marker will be offset by a number equal to the height of the marker multiplied by this value.
+   * @param typeAnchorX The x-axis position type of a marker. See POS_* flags in MarkerLayout.
```
moagrius: instead of calling this `typeAnchorX` can we try something like `xAnchorBehavior`?  same for `y` equivalents...

tileview/src/main/java/com/qozix/tileview/markers/MarkerLayout.java
```diff
@@ -16,6 +16,10 @@
 
   private MarkerTapListener mMarkerTapListener;
 
+  public static final int POS_RELATIVE = 0;               // Position is relative
```
moagrius: the names of these flags aren't immediately clear to me.  maybe something like `ANCHOR_BEHAVIOR_RELATIVE` and `ANCHOR_BEHAVIOR_ABSOLUTE`?

tileview/src/main/java/com/qozix/tileview/markers/MarkerLayout.java
```diff
@@ -16,6 +16,10 @@
...
+  public static final int POS_ABSOLUTE_LEFT_TOP = 1;      // Position is absolute calculated from left or top of marker
+  public static final int POS_ABSOLUTE_RIGHT_BOTTOM = 2;  // Position is absolute calculated from right or bottom of marker
```
moagrius: have you found it's necessary to have both the top/left and bottom/right flags?  it seems a little cleaner to _just_ have relative vs absolute...

tileview/src/main/java/com/qozix/tileview/markers/MarkerLayout.java
```diff
@@ -120,15 +138,25 @@ protected void onMeasure( int widthMeasureSpec, int heightMeasureSpec ) {
...
         int actualHeight = child.getMeasuredHeight();
-        // offset dimensions by anchor values
+        // get anchor offsets
+        float widthMultiplier = (layoutParams.anchorX == null) ? mAnchorX : layoutParams.anchorX;
+        float heightMultiplier = (layoutParams.anchorY == null) ? mAnchorY : layoutParams.anchorY;
+        // offset dimensions by anchor values (default typeAnchor = 0)
         float widthOffset = actualWidth * widthMultiplier;
         float heightOffset = actualHeight * heightMultiplier;
+        // handle typeAnchorX options other than 0
+        switch (layoutParams.typeAnchorX) {
```
moagrius: no need to recompute what was already figured on lines 148 and 149, may as well add that as the `default` case to your switch statements

tileview/src/main/java/com/qozix/tileview/markers/MarkerLayout.java
```diff
@@ -268,6 +302,16 @@ public LayoutParams( int width, int height, int left, int top, Float anchorLeft,
...
     }
 
+    public LayoutParams( int width, int height, int left, int top, Float anchorLeft, Float anchorTop, int typeAnchorX, int typeAnchorY ) {
+      super( width, height );
+      x = left;
+      y = top;
+      anchorX = anchorLeft;
+      anchorY = anchorTop;
+      this.typeAnchorX = typeAnchorX;
+      this.typeAnchorY = typeAnchorY;
```
moagrius: no `this` on the above 2 lines.  when the member names are updated to include the `m` prefix, it will naturally disambiguate

tileview/src/main/java/com/qozix/tileview/markers/MarkerLayout.java
```diff
@@ -199,6 +227,12 @@ protected boolean checkLayoutParams( ViewGroup.LayoutParams layoutParams ) {
      */
     public Float anchorY = null;
 
+    /**
+     * Flag to determine if anchorX/Y are relative or absolute values
+     */
+    public int typeAnchorX = 0;
+    public int typeAnchorY = 0;
```
moagrius: please follow the convention where non-public member variables are prefixed with `m`

tileview/src/main/java/com/qozix/tileview/widgets/ZoomPanLayout.java
```diff
@@ -39,6 +39,7 @@
   private int mBaseHeight;
   private int mScaledWidth;
   private int mScaledHeight;
+  private int mExtraBorder = 0;
```
moagrius: default value is 0, no need to set it

tileview/src/main/java/com/qozix/tileview/widgets/ZoomPanLayout.java
```diff
@@ -196,6 +197,15 @@ public void setSize( int width, int height ) {
     requestLayout();
   }
 
+  /**
+   * Sets the size of an extended border around the view.
+   *
+   * @param border  Extra size of the scaled view.
+   */
+  public void setExtraBorder(int border) {
+    mExtraBorder = border;
```
moagrius: this isn't really a "border", it's more like padding.  maybe call it something like "inset" or "visual padding"?

tileview/src/main/java/com/qozix/tileview/widgets/ZoomPanLayout.java
```diff
@@ -598,19 +608,19 @@ protected int getConstrainedScrollY( int y ) {
   }
 
   protected int getScrollLimitX() {
-    return mScaledWidth - getWidth();
+    return mScaledWidth - getWidth() + FloatMathHelper.scale(mExtraBorder, mScale);
```
moagrius: let's precompute these values.  i'd probably do something like have `mExtraBorder` and `mScaledExtraBorder`.  whenever either value changes (scale or the border) recompute it and save the value in the latter variable.  that way each of these 4 methods (`getScrollLimitX`, `getScrollLimitY`, `getScrollMinX `, `getScrollMinY `) do not need to recompute every time and can just reference the `mScaledExtraBorder` value (or whatever name that turns out to be, in consideration of other comments)

tileview/src/main/java/com/qozix/tileview/TileView.java
Derpalus: Reasonable, I'll rename them.

tileview/src/main/java/com/qozix/tileview/markers/MarkerLayout.java
Derpalus: I'll rename these too.

tileview/src/main/java/com/qozix/tileview/markers/MarkerLayout.java
Derpalus: The reason I had this was that if you have for example a 9-patch and you want the center to be 50 pixels in from the right then there is no way to do this counting from the left as you don't know the width of the marker until it's actually displayed. Hence you need one flag from each direction.

However one could also choose to have a negative offset meaning it's calculated from the left/top and a positive offset meaning it's calculated from the right/bottom. A method like that however would prevent you from setting an absolute position outside the marker as an anchor point.

The current relative offsets are counted from the top left I think and they are always negative if you want the anchor to end up inside the bounds of the marker (ie. shifting the marker up or to the left in relation to the anchor point).

Which method would you prefer?

tileview/src/main/java/com/qozix/tileview/markers/MarkerLayout.java
Derpalus: Reasonable, I'll fix that.

tileview/src/main/java/com/qozix/tileview/markers/MarkerLayout.java
Derpalus: They are actually public. But I will rename them to have better names and remove the zero initialization.

tileview/src/main/java/com/qozix/tileview/markers/MarkerLayout.java
Derpalus: I have renamed the variables so this is no longer needed.

tileview/src/main/java/com/qozix/tileview/widgets/ZoomPanLayout.java
Derpalus: I'll fix that.

tileview/src/main/java/com/qozix/tileview/widgets/ZoomPanLayout.java
Derpalus: Agreed. Naming has never been my strong suite ;)
I'll call it `setImagePadding` instead.

tileview/src/main/java/com/qozix/tileview/widgets/ZoomPanLayout.java
Derpalus: Sounds like a good idea. I'll fix that.

How about `getConstrainedScrollX` and `getConstrainedScrollY`? Couldn't those have more precomputing as well with the same reasoning? (though that is outside the scope of these changes)

tileview/src/main/java/com/qozix/tileview/markers/MarkerLayout.java
moagrius: it's late and i may not be thinking this through appropriately, but what if the relative and absolute values were combined?  e.g.,

1.0 relative and 10px offset = 10px from bottom right
0.0 relative and 10px offset = 10px from top left
-1.0 relative and 10px offset = (width + 10px from left, height + 10px from top)
-0.5 x relative and 0px x offset, -1.0 y relative and -20px y offset = would be centered horizontally, and 20px above the point...

etc...

thoughts?

tileview/src/main/java/com/qozix/tileview/markers/MarkerLayout.java
moagrius: oh you're right about the `public` access, no need for the `m` prefix

tileview/src/main/java/com/qozix/tileview/widgets/ZoomPanLayout.java
moagrius: That's a good point.  TBH it's been many years since that was written so it might have just been an oversight or laziness, or maybe it's because there's so many contributing variables that it seemed easier to compute on the fly...  this library is old and rickety, i'll admit (thus, V3)

tileview/src/main/java/com/qozix/tileview/markers/MarkerLayout.java
Derpalus: That... is a very good idea. Simply get rid of the flags and have both a relative position and an absolute position for the anchor point. I'll fix that and submit again.

tileview/src/main/java/com/qozix/tileview/markers/MarkerLayout.java
```diff
@@ -16,6 +16,10 @@
 
   private MarkerTapListener mMarkerTapListener;
 
+  public static final int ANCHOR_BEHAVIOUR_RELATIVE = 0;               // Position is relative
+  public static final int ANCHOR_BEHAVIOUR_ABSOLUTE_LEFT_TOP = 1;      // Position is absolute calculated from left or top of marker
+  public static final int ANCHOR_BEHAVIOUR_ABSOLUTE_RIGHT_BOTTOM = 2;  // Position is absolute calculated from right or bottom of marker
+
```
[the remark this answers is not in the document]
Derpalus: I missed that? Dang.
Well it's fixed now.

moagrius:
aiming for this weekend to merge

Derpalus:
And there was much rejoicing! :D

moagrius:
Ha - i believe 2.2.9 is now on jcenter - testing out some demos to make sure there weren't any unexpected side effects

moagrius:
after a quick run through, everything looks good (as expected).  feel free to add samples to the demo, instructions to the readme, or a page to the wiki, but don't feel obligated.

thanks for contributing!

PR #475, opened 2018-05-04
Replacement of "compile" with "implementation"

TurboProgramming:
As "compile" is going to be deprecated end-2018, this PR replaces with "implementation" that is the new usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Javadocs are [here](http://moagrius.github.io/TileView/index.html?com/qozix/tile
 ### Installation
 Gradle:
 ```
-compile 'com.qozix:tileview:2.2.9'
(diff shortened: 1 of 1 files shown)


peterLaurence:
Btw, this could be extended to [tileview/build.gradle](https://github.com/moagrius/TileView/blob/master/tileview/build.gradle), where two `compile` statements can be replaced.
I've done this on other projects recently, it works.

TurboProgramming:
Completely forgot those, those need to be changed too.

TurboProgramming:
Done!

moagrius:
i don't see any new commits?

peterLaurence:
@TurboProgramming you pushed inside your `master` branch instead of  `patch-1`. That's why we don't see the new changes.

PR #489, opened 2018-09-30
Feature/rotatable map

dilara91:
This branch will enable rotations. Calculates scroll limits and tiles according to rotation. The sample code is on the RealMapTileViewActivity.java in the demo repo. That code will rotate the map with 30 degrees every time a marker is tapped.

Basically all you have to do is setting the map rotatable first and then giving the rotation degrees.

      tileView.setRotational(true);

      tileView.rotate(45);

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,9 +3,10 @@
 buildscript {
     repositories {
         jcenter()
+        google()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.2.1'
+        classpath 'com.android.tools.build:gradle:3.0.1'
 
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
@@ -15,6 +16,7 @@ buildscript {
 allprojects {
     repositories {
         jcenter()
+        google()
     }
 }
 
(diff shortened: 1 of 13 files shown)


moagrius:
Wow - very cool - sorry for the delay in responding - I'll try to check this out this weekend - thanks for the contribution

moagrius:
so this is very cool, but considering that it's on version 2, which is no longer supported, i'm not sure what to do with it - probably leave it as a fork?  as far as migrating to version 3, that would be cool but i'd definitely want it as a plugin rather than modifications to the core - do you believe that's possible?

dilara91:
@moagrius Hi. Yeah leaving as a fork seems to be more suitable. If people still use the version 2 they can get it from the fork. 
As for the version 3, I haven't looked at it yet so I dont know. Still would like to do it sometime when I have spare time. Just started my PhD so it probably will not be soon. No promises. If anyone is willing that would be great.

moagrius:
Sounds reasonable, thanks!

srbhagwatsrb:
@moagrius Does version 3 include this feature?

moagrius:
@srbhagwatsrb no.  as i've mentioned, i have no intention of adding rotation.  if someone else creates a PR that somehow manages to add rotation without breaking any public APIs, i'd include it, but i find that unlikely.

Issue #494, opened 2018-11-22
Documentation slight outdated (for quick setup)


perrochon:
Tried to get this running. Looks super nice and promising. A few things I noticed while setting it up.

1. I needed to set min sdk to 26 (Oreo) to compile. Not a problem, and easily fixable through error messages

2. Could it be that .defineZoomLevel replaced .defineDetail and .addDetailLevel? readme.md refers to defineDetail in the quick setup , and the wiki uses addDetailLevel. The demo code shows how.

moagrius:
yeah the javadocs are completely outdated and i haven't gone through all the wiki pages.  the demo and readme should be relatively correct, although feel free to fork and PR (or just edit the wiki directly) if you have the bandwidth to correct any typos or misnamed methods.  otherwise i'll get to it when i can - i usually work in short bursts every few months.

yeah it's `defineZoomLevel(int zoom, Object data)` now.  https://github.com/moagrius/TileView/blob/master/tileview/src/main/java/com/qozix/tileview/TileView.java

moagrius:
you should be able to edit the wiki directly, but it's np i can do it - what page/line?

Pavle37:
I also get an error message for min sdk to be 26. How can i fix this to use it with min sdk 19?

moagrius:
This should be in a different issue thread, but nity demo and library have minSdk of 19.  Check other Gradle or manifest settings in your project, or other dependencies

tva-TIS:
Also in README.md `TileView.Builder().build()` does not return the build TileView anymore but void and takes the TileView to build as a paramter.

moagrius:
i will fix this as soon as i get some time for this project (weeks away)

PR #496, opened 2018-11-24
Update README.md

perrochon:
.defineZoomLevel replaced .defineDetail

As discussed in issue https://github.com/moagrius/TileView/issues/494


README.md
```diff
@@ -40,7 +40,7 @@ protected void onCreate(Bundle savedInstanceState) {
   super.onCreate(savedInstanceState);
   TileView tileView = new TileView.Builder(this)
   	.setSize(3000, 3000)
-  	.defineDetail("tile-%d-%d.png")
+    .defineZoomLevel("tile-%1$d_%2$d.png")
```
moagrius: whitespace.  also can we use the original template? (less confusing with a single separator instead of two different ones, and the format indices aren't necessary)

PR #516, opened 2019-05-15
limit zoom to minimum of 0 when computing

moagrius:
@peterLaurence if you don't have time or bandwidth to review this one, it's pretty straightfoward, and i was able to repro the bug, and visually experience that this fix does indeed correct it, so i can merge.  that said, another pair of eyes confirming that is always a good thing, so i'll leave it as an eligible PR for now.

also, the `setMaximumScale` needs to be in `ScalingScrollView` which is another repo now, so I'll make a PR there following this.

thanks!


peterLaurence:
+1, works for me

Issue #511, opened 2019-05-11
Restore position and zoom upon configuration change


peterLaurence:
Hey Mike,

So I finally started to upgrade from v2. And boy it's harder than it should be. One of the reasons is that the `BitmapProvider` type leaked into the core of my business logic (so some classes were coupled with). Anyway, i'm fixing this so TileView-related stuff should be more self-contained.

And I started to see some issues here and there. I will create issues for each one of course.
So this one is pretty much self explanatory, the state save-restore was something we added in v2. 
I can do PR. But for instance I have a lot of things to finish on my app.

moagrius:
Ok let me know if I can help.  Obviously any issue you create will get addressed immediately

moagrius:
i'll get to this next.  i'm out tomorrow but should be able to get started thursday.  there's also the wierd scaling past limits thing reported by another user i need to do.

moagrius:
For documentary purposes (mostly to remind myself).  This has been bandaid-ed by suppressing config changes on the activities, but this still needs to be fixed properly with good save and restore instance state logic, which is complicated by the `prepare` and `Builder` mechanisms (I think).  Both the marker and path plugins all apart currently - I think scroll and scale is fairly straightforward.  Note that restore instance state happens well after `Activity.onCreate`.  I suspect it happens before `TileView.onReady` but am not sure - Probably depends on if there's a diskcache or not.

This is the most important (and possibly last?) outstanding non-trivial bug.  I'm targeting Monday, June 10th to deliver a complete fix.

moagrius:
fixed in 4.0.5

PR #521, opened 2019-05-20
Issue 510

moagrius:
@peterLaurence not a huge rush on this one, I still need to get #511 done, and you're welcome to wait on this one (510) until 511 is up as well (or you can do this one now that's more convenient for you - your call).  

TBH I'm not 100% sure what solved this, I just kept trying things, and TBF the remote version on an emulator is very very slow, but on a real device it's pretty snappy and I can't get it to ever miss a tile.  I'm inclined to close, despite funky emulator behavior sometimes (it's always slow, but only very rarely misses tiles).

### DIFF
--- a/demo/src/main/java/com/moagrius/TileViewDemoAdvanced.java
+++ b/demo/src/main/java/com/moagrius/TileViewDemoAdvanced.java
@@ -58,7 +58,7 @@ public class TileViewDemoAdvanced extends Activity {
     TileView tileView = findViewById(R.id.tileview);
     tileView.setScaleLimits(0, 2f);
     new TileView.Builder(tileView)
-        .setSize(16384, 13312)
+        .setSize(16384, 13056)
         .defineZoomLevel("tiles/phi-1000000-%1$d_%2$d.jpg")
         .defineZoomLevel(1, "tiles/phi-500000-%1$d_%2$d.jpg")
         .defineZoomLevel(2, "tiles/phi-250000-%1$d_%2$d.jpg")
--- a/demo/src/main/java/com/moagrius/TileViewDemoAdvanced.java
+++ b/demo/src/main/java/com/moagrius/TileViewDemoAdvanced.java
@@ -55,9 +55,6 @@ public class TileViewDemoAdvanced extends Activity {
   protected void onCreate(@Nullable Bundle savedInstanceState) {
     super.onCreate(savedInstanceState);
     setContentView(R.layout.activity_demos_tileview);
-    BitmapFactory.Options options = new BitmapFactory.Options();
-    options.inPreferredConfig = Bitmap.Config.RGB_565;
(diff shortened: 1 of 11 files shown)


peterLaurence:
Ok, will try that tomorrow.

moagrius:
Thanks!  Working on save/restore as we speak...

peterLaurence:
So my testing didn't reveal any gaps. But it doesn't really mean it won't happen in some scenario.
We know it happened (rarely) in v2. 
Looking at your changes, I can see that you introduced a retry (a single one). The probably of having a tile decoding failing twice in a row is very low, but I'm wondering if we should set a limit to retries.
Also, before attempting a another tile decoding, should we check if the tile is still visible in the viewport?

moagrius:
> So my testing didn't reveal any gaps. But it doesn't really mean it won't happen in some scenario.
We know it happened (rarely) in v2.

That's a very good point and I don't want to pretend that it's perfect, but I never get gaps in a real device.  I _think_ I've seen a couple missing pieces on an emulator.  It might be worth trying that whole "unfilled area" thing again...  I'm not sure if the Tile thinks it's decoded but isn't rendering, or the bitmap is null, or that it knows it's not decoded.  Once I get the hot issues out of the way, I'll spend some time doing deep investigation on this (like hotspots over every tile so I can tap and check the state - but that's a pretty heavy lift).

> I'm wondering if we should set a limit to retries.

so pass an int indicating the number of times it should retry?

> Also, before attempting a another tile decoding, should we check if the tile is still visible in the viewport?

That's a very good idea, I'll include that.

peterLaurence:
> so pass an int indicating the number of times it should retry?

I was thinking of removing the limit of retries, provided the tile is still in the viewport. But it's just a suggestion.

moagrius:
Gotcha.  I think we need some kind of limit in case of missing files or 404s, which would create an infinite loop

peterLaurence:
You're right. I had more in mind tiles from assets or sd card.

moagrius:
So yeah maybe we let them pass in a limit?

peterLaurence:
With a default of 1 or 2, yes passing a limit could be interesting.

moagrius:
will do

moagrius:
so i'm going to make the updates and merge and don't think i need you to re-review.  i have a nutty weekend so it'll probably be early next week.  if you could check out #511 and see if you agree that works as a temporary workaround, i'd love to merge that into this, before merging this into master.  if not, it's not the end of the world.

fwiw, just in case, i had the tiles stored in google cloud but under my work email, which was linked to my employers account.  i deleted that project and am re-uploading to a persona email, but the urls for the remote demo may fail in the branch.  you probably wouldn't even have noticed, but i figured worth a mention.

thanks as always

peterLaurence:
Ok omw to test it.

PR #524, opened 2019-05-28
Save and restore instance state properly

moagrius:
@peterLaurence wow this was a bear!  This addressed #511.  It seems to work well, but it was extremely complicated.  I think there are several standalone "fixes", but I wanted to be thorough and catch anything else that might come up, so there are multiple code updates that should resolve the issue we were seeing.

This also fixes on the center point of the screen (as you put it, "area of interest"), so it should no longer show the "walking" behavior you described before.

Plugins now work as well - check the Advanced demo to see markers and paths are always rendered properly.

This also removes some buggy code, and big bug where a Handler continues to fire delayed messages continuously through the life of the app, the number increasing with each call to layout.  If this is approved, I'll make a note in the readme for all 4.0.0 - 4.0.4 users to update to this (4.0.5) immediately.

There's a new bug but one I think is fairly trivial, and I can sit on while I finish the book - if you scale larger than 1f, the scroll position is off visually during save and restore.  The math should be the same, but it definitely happens.  Also when scaling past 1f, markers vanish 🤷‍♂ I'll create a new ticket, once this PR is merged


peterLaurence:
I can feel your pain :)
I will test this tonight and let you know.

moagrius:
Thanks!!

peterLaurence:
Approved!
+1

moagrius:
woooooooooo!

thank you sir!  i think i should be able to leave you in peace for a while...

now back to the book ;)

our final draft for all tech review is due July 1.  You can either continue with the half you have (if you do want to continue with what you have, I can send you an updated PDF with a few changes), or you can wait for the whole thing (probably just before the deadline of July 1) and I can send you the whole book to do in one fell swoop.  The first half will require much more technical review than the second half, but it's totally up to you.  LMK - no rush!

peterLaurence:
For now, I have reviewed chapter 1 to 8. If you feel that I should wait for the final draft to continue, no problem. Otherwise, I can continue with an updated copy.

moagrius:
Cool, sending

Issue #409, opened 2017-04-29
Scaling markers


sdargutev:
Hello,
thank you for the awesome library! 
I wonder is there a way to add markers that also scale? Currently markers remain always the same size. I can call marker.setScaleX(scale) and marker.setScaleY(scale) in the onScaleChanged(...) method, but the problem is that in this case the MarkerTapListener does not detect the correct tap location after the scaling.  I am guessing, it does not update marker locations as they scale in this case..

Best regards,
S. Dargutev

moagrius:
You're correct, in and Android View system, scaling a View will not affect it's layout or hit area.  You can manage those manually, or use `HotSpots` at those areas.  In version 3 this might be a good candidate for a plugin.

ameron32:
I was trying to produce Markers that scale up and down as well. For anyone looking for a functional solution prior to the version 3 plugin implementation, here is what worked for me.

If you copy/paste the code from MarkerLayout into a new class, say ScalingMarkerLayout, you can create a layout that works. There are only 2 lines of code to change in your new ScalingMarkerLayout class in order to achieve what you want from scaled markers.

from:
```
  @Override
  protected void onMeasure( int widthMeasureSpec, int heightMeasureSpec ) {
        ...
        // actual sizes of children
        int actualWidth = child.getMeasuredWidth();
        int actualHeight = child.getMeasuredHeight();
        ...
}
```

to:
```
  @Override
  protected void onMeasure( int widthMeasureSpec, int heightMeasureSpec ) {
        ...
        // actual sizes of children
        int actualWidth = FloatMathHelper.scale(child.getMeasuredWidth(), mScale);
        int actualHeight = FloatMathHelper.scale(child.getMeasuredHeight(), mScale);
        ...
  }
```

This forces the Markers to scale in the same manner than everything else does. Seems to work perfectly for drawing and touch capture (via the processHit() method) for scaled markers. You'll have to add the ScalingMarkerLayout to your TileView since this is not automatic. If you extend TileView you can add it to your constructor...

```
  public SuperTileView(Context context, AttributeSet attrs, int defStyleAttr) {
    super(context, attrs, defStyleAttr);
    ...
    ScalingMarkerLayout scalingMarkerLayout = new ScalingMarkerLayout(context);
    addView(scalingMarkerLayout, 3); // above mScalingLayout & below mMarkerLayout
    ...
  }
```
...or to your Activity, Fragment, or ViewGroup, anywhere you have instantiated your TileView...
```
[...]

PR #536, opened 2019-12-13
Added Scaling Marker Plugin #409

yashx:
I need scaling markers for a project and saw a feature request for it #409  so here it is.
I also added the option to decide the scale when the original dimensions should be shown.


moagrius:
NIce, thanks

